PR #1214, opened 2015-10-07
Webshop (WIP)

cXhristian:
Obvious things that are missing:

- [ ] Dashboard admin (currently using django admin)
- [ ] Design that doesn't suck
- [ ] Stripe


cXhristian:
This will have to work for now.

PR #1099, opened 2015-03-12
Dashboard events (WIP)

melwil:
Finishing up work on the events dashboard

- [x] Create Events
- [x] Add and manage AttendanceEvent
- [ ] Add and manage Reservations
- [ ] Show reserved seat count in attendee page
- [ ] Small bug where toggling doesn't work after tables are redrawn in attendees
- [ ] Wait for datetimewidget to come out with a new version after 0.9.3
- [ ] Image handling

PR #1251, opened 2015-10-18
Dashboard for authentication app

sklirg:
Initial dashboard app for `apps.authentication`.

screenshot 2015-10-18 05 50 36

ToDo:
- [x] Lists
- [x] View profile
- [x] Edit profile
- [x] Searching


frecar:
Please rebase with master branch to make this work with Frigg

sklirg:
@frecar We don't have master :trollface: 

I'll fix it.

hernil:
"NoReverseMatch at /dashboard/auth/user/X/edit"

I get this when trying to edit a user from the dashboard.

hernil:
Also, did you look into using watson for search?

myth:
@frecar seems like there was a connection reset by peer during the tests. Have any input? Also, seems like the SSL package used by request is missing.

sklirg:
Interesting, I'll have a look at it.

frecar:
@myth: Please try to add requests[security]==2.8.0 to requirements.txt

frecar:
You should also upgrade to latest version of django_extensions(1.5.9), to avoid the problem with south_migrations and django 1.8.6

frecar:
Great, the only thing missing now is to fix all pep8/flake8 issues in the project :)

sklirg:
@frecar Yes! And before we can do that, we have to merge the WIP branch #1099. So we'll have to live with some "failed" builds for now, or up the acceptable error limit so we can see the beautiful green checkmarks until that branch is merged. ;)

frecar:
Okey, I guess you have control! ;) 

However, I would not recommend to change the limit, but fix the errors asap instead!

cXhristian:
Yeah, let's not change the limit. That's how we ended up with this mess to begin with.

hernil:
I agree on not raising the limit. There's also a couple of other issues regarding complexity that we should fix as well.

sklirg:
I totally agree, however, the branch is not ready for merge yet so we cannot merge it, therefore we can't PEP8 it.

Issue #1221, opened 2015-10-14
Adding fields to user profile


nornes:
The user-profiles looks really good, but I'd love to see fields to links to Github/LinkedIn etc.

PR #1275, opened 2015-10-21
Adds website fields to user profile

plusk:
Adds #1221


niklasmh:
Would be nice if the url's were links too. Also, if the "https://" protocol is missing on github links the form could assume the user has filled in the username. Links on linkedin are a bit more complicated if you don't have a pre-generated link.

sklirg:
This PR is closed. Create a new issue with your input @niklasmh :)

PR #1281, opened 2015-10-21
Split requirements into several files

### LINKED ISSUE #1260: Separate requirements in develop, testing and production
We should separate requirements for various environments. What sparked this was requiring some postgresql magic to install all requirements (`psycopg2`), and e.g. moonshine does not want postgres just for tests to pass (or run, for that matter).

Requirements that should be removed from `requirements.txt` and added to `requirements-<env>.txt`:

#### Production requirements [src](https://github.com/dotKom/onlineweb4/blob/develop/requirements.txt#L9):
- `psycopg2` (yes, some people use postgres in development. However, example-local.py uses sqlite and most people will therefore not need this package. *1)

#### Testing requirements [src](https://github.com/dotKom/onlineweb4/blob/develop/requirements.txt#L38):
- `cov-core`
- `coverage`
- `django-nose`
- `factory-boy`
- `lettuce`
- `nose` *2
- `nose-cov`
- `teamcity-messages`

#### Development requirements (these can be/should be fully optional)
- `django-debug-toolbar` (it's not in `requirements.txt` as of now, but maybe it should? [wiki](https://github.com/dotKom/onlineweb4/wiki/Django-debug-toolbar))

*1: It also fails to install and stops requirements.txt installation if people don't have postgresql dev headers ([ex.](https://online.ntnu.no/tc/viewLog.html?tab=buildLog&buildTypeId=Onlineweb4_Build&buildId=29))
*2: If we remove this one we have to change default runner class in base settings as the current testrunner is nose.

sklirg:
Fix #1260

Keep nose installed for local testing purposes

Require memcache

Don't load gunicorn app in ow4


frecar:
Upgrade django_extensions to avoid the South problem :)

frecar:
and add requests[security] to the requirements

hernil:
Would you mind updating the prodsetting script on nansen to account for multiple requirements files?

sklirg:
Done.

sklirg:
Merged with develop.

PR #1305, opened 2015-10-28
Uses unidecode for slugs instead of slugify

### LINKED ISSUE #1304: Replace norwegian characters in slugs
We actually should do a replace on æøå on titles before sluggifying them. It's more readable than just removing the characters.

andelar96:
Fixes #1304


hernil:
Unidecode should be wrapped around slugify as to strip the string of special characters that are ascii compatible. Example: `slugify(unidecode(bundle.data['title']))`

PR #1341, opened 2015-11-05
Bump django extensions and add requests security for frigg

sklirg:
Since we use `django>=1.8.0,<1.9.0` as a requirement we automatically get the newest django version if there's a new one. While not everyone installs requirements every time they start to work on the project, our deployment script will.

If there's going to be a release today, this patch is required as Django 1.8.6 (released yesterday) disallows south migrations in the same directory as django migrations, something django_extensions does in v. 1.5.6 (pointed out by @frecar in #1251 and #1281).

This patch bumps django extensions to a version where this is not a problem any more (v 1.5.9), and it adds some requirements for frigg tests. (while both the aforementioned PRs have included these requirements bumps, these requirements should be in develop ASAP and not having to wait for those feature branches to be merged.)


hernil:
I ran into this myself today. Well spotted!

Issue #1080, opened 2015-02-26
Update waiting list bump mail if event has payment


Torrib:
Add information notifying the user about the required payment when he's bumped from waiting list.
Should contain information about the sum and deadline (deadline should be 48 hours if deadline < 48h)

PR #1342, opened 2015-11-06
#1080 is now fixed like a pro 8|

Niklasmh:
Det havner en liten padding-left på body (17px), ellers er problemet løst. Mulig problemet kan forsvinne med senere oppdateringer av bootstrap.


OptimusCrime:
Please delete file `setuptools-18.4.zip`.

Niklasmh:
Frigg acts unfriendly when it comes to testing. Can't figure out why it fails. Does anyone find something suspicious?

image

myth:
Well, until we have removed all Flake8 warnings Frigg will still mark the build as a failure.
Although that particular error is caused by the build not being executed in a proper environment.
The entire Django module is missing in that environment. You could attempt to perform a dummy-commit and see if the rebuild also fails. If so, there is an error with the Frigg environment. If it succeeds, it was most likely a one-time error.

OptimusCrime:
In other words, not your fault.

frecar:
Frigg fails because you are working on an old branch (without frigg.yml and a proper tox.ini)

templates/events/details.html
```diff
@@ -189,7 +189,24 @@ <h2 id="event-details-heading">
                                     <p class="status-text">{{ user_status.message }}</p>
                                 {% endif %}
                             {% else %}
-                                <p class="status-text">{{ user_status.message }}</p>
+                                <p class="status-text">{{ user_status.message }}
+									<br><button type="button" class="btn btn-primary" data-toggle="modal" data-target="#myModal">Info</button></p>
+									<div id="myModal" class="modal fade" role="dialog" style="color:#666;">
```
lizter: Inline styles

Niklasmh: I was in a hurry when I tried to implement the solution and I may have overlooked the naming. The inline-style was a quick-fix for the text color (is there an existing class for text colors?). I will try to make it look a bit more professional than this.

PR #1343, opened 2015-11-10
Bump feedme to fix bug in 1.8.6

sklirg:


sklirg:
I'll merge this so we can eat something tomorrow.

PR #1345, opened 2015-11-12
Save Stripe ID for webshop purchases, more informative receipt emails

sklirg:
Basically I wanted to add the stripe charge id so a customer has some reference to easily match a given transaction (using the uid rather than filtering on user...) if there are any questions, or for a refund. This information should be saved on our servers for easy retrieval, rather than we having to query Stripe to get this information.

Tbh, it's a really important fix and it should be released ASAP, unless anyone have any problems with the implementation.

I also included a simple list of what was purchased.

### DIFF
--- a/apps/payment/views.py
+++ b/apps/payment/views.py
@@ -139,10 +139,11 @@ def webshop_pay(request):
                 )
 
                 order_line.pay()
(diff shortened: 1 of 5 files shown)


cXhristian:
Good work!

PR #1350, opened 2015-11-15
Add forgotten comment field to poster model

### LINKED ISSUE #1349: Posters comments
Currently posters comments are not being saved. 

Follow-ups to the order should be sent as emails. We should include the comment in the original email.

nixolas1:
Resolves #1349.

And added field to admin.py

### DIFF
--- a/apps/posters/models.py
+++ b/apps/posters/models.py
@@ -41,6 +41,7 @@ class OrderMixin(models.Model):
         null=True
     )
     description = models.TextField(_(u"beskrivelse"), max_length=1000, blank=True, null=True)
+    comments = models.TextField(_(u"kommentar"), max_length=1000, blank=True, null=True)
     amount = models.IntegerField(_(u'antall opplag'), blank=True, null=True)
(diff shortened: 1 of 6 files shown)


apps/posters/models.py
```diff
@@ -41,6 +41,7 @@ class OrderMixin(models.Model):
         null=True
     )
     description = models.TextField(_(u"beskrivelse"), max_length=1000, blank=True, null=True)
+    comments = models.TextField(_(u"kommentar"), max_length=1000, blank=True, null=True)
```
sklirg: TextFields do not require a `max_length` attribute - the only thing this will do here is to limit the input text area to 1000 characters. If that's not the desired effect, just strip it from the model.

OptimusCrime: Why have a max length at all? I suggest removing it from description too. It can never be too much information.

sklirg: :+1:

hernil: Update this for merge

PR #1355, opened 2015-11-17
Featured events by registration start

### LINKED ISSUE #1285: Sort events on front page by registration start rather than event start
This way we won't have problems with events having registration before even being on the front page and the listing will actually be based on when one has to actually do something with regards to an event. 

I can't really remember whether this was ever brought up as an option when designing the site ( @lizter, @melwil ). Personally I like the idea, but I'd like some input.

sklirg:
Fixes #1285.

Committed code is well documented with the changes.

This new manager will order events by registration start if it's an attendance event and attendance_event.registration_start is within a time delta of 7 days.

### DIFF
--- a/apps/events/models.py
+++ b/apps/events/models.py
@@ -39,8 +39,10 @@ class EventOrderedByRegistration(models.Manager):
     Order events by registration start if registration start is within 7 days of today.
     """
     def get_queryset(self):
-        week_back = timezone.now() - timedelta(days=7)
-        week_in_future = timezone.now() + timedelta(days=7)
+        DELTA_FUTURE = settings.OW4_SETTINGS.get('events').get('OW4_EVENTS_FEATURED_DAYS_FUTURE')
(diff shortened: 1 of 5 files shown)


hernil:
I like the solution :-) However, what would happen with "ekskursjonen"? Would this hover on the front for a month or more waiting for the registration window to close? Might be useful to have a "timeout".

sklirg:
Yes it would. Even worse for Åre, which opens in October and closes in December.

We could sort on `registration_start` for the first three days or so, and then event_start (like normal)?

hernil:
Yep, that would be a good solution. A week seems reasonable, but if you want to be really fancy about it it could be a week or three days depending on how many events are queued for that week.

sklirg:
I've come up with a solution which solves the aforementioned problems whilst also fulfilling the initial request. Have a look!

I also included some test cases.

cXhristian:
:shipit: :+1:

hernil:
+10 for the testing!

PR #1384, opened 2016-01-20
Python 3

### LINKED ISSUE #1334: Upgrade Onlineweb to utilize Python 3.
Me and Duvvie the Dove @cXhristian were up late at the office tonight, and came to discuss how much would be required to bump Onlineweb to use Python 3. It started when discussing the unicode decode stacktrace error from password reset that came in during the night. I mentioned the `2to3` tool used to transition python2 applications to python3, which we investigated.

After some experimentation, dependency bumping, and test fixing (division rounding errors due to division type presedence), @cXhristian managed to get a working version of Onlineweb up and running on Python3 in about 10 minutes.

There were still some issues with coverage, xcov or cobertura (can't remember which), but when those are resolved I think we should transition to python 3 as soon as possible.

Not only would we no longer have to account for unicode related issues, but also be more prepared to utilize newer versions of python that introduce awesome things like [async/await](https://www.python.org/dev/peps/pep-0492/) and type hinting.

cXhristian:
Fixes #1334 
:balloon:


cXhristian:
Only flake8 fails now which it does on python2 too.

Issue #1379, opened 2016-01-08
The same AttendanceEvent status_codes is used multiple times


OptimusCrime:
Kind of defeats the porpupseoa012+

https://github.com/dotKom/onlineweb4/blob/develop/apps/events/models.py#L594
https://github.com/dotKom/onlineweb4/blob/develop/apps/events/models.py#L604
https://github.com/dotKom/onlineweb4/blob/develop/apps/events/models.py#L612

sklirg:
Regarding your comment on the referenced issue (#1378), reworking the whole
template, we should make sure to follow the standards for status codes.
https://en.m.wikipedia.org/wiki/List_of_HTTP_status_codes

None of the current errors are server side errors, thus they should not be
5xx. 400 bad request is a nice catch-all. 403 permission denied works for a
couple of use cases. There won't be a status code for everything, so we
should probably also look at re-defining the existing ones and combine them
where that makes sense.

OptimusCrime:
@sklirg : They are not HTTP status codes, they are just internal codes @melwil used to simplify the feedback. I dont think there is a HTTP code that signals "du har pikk og har fått utsatt påmelding i 24 timer". Also, we should not mess with the HTTP status codes that Django sends, as every page load is in fact a 200 Ok response. If we start returning 403, Google and others will act weird on us.

myth:
Googlebot does not perform POST requests or any ajax shenanigans, so they can safely be used for that purpose. But ofc, a regular GET on the event should never return anything other than 200, unless it is a restricted event and the user is missing appropriate permissions.

OptimusCrime:
We're talking about the event view, shich is a regular GET.

sklirg:
My bad, I guessed they were HTTP status codes because they were 4xx and 5xx.

---

What are they used for? A quick search for 501 and 503 contains no results for anything other than the source code in `models.py`.

If they should still be included, I think they should be renamed as to not conflict with the HTTP status code name. Django HttpResponse already has `HttpResponse.status_code`, adding a `status_code` entry to the context dictionary is confusing.

---

Every _page load_ should be 200 OK, I agree. When we do ajax requests _on_ the page, thus not requesting a page, but rather some [async request](https://github.com/dotKom/onlineweb4/blob/6dea161fb6d1d70ba7dd203823456532355c46d8/apps/events/views.py#L70) returning some data, returning non-200 is perfectly fine.

OptimusCrime:
[...]

Issue #1378, opened 2016-01-08
Event opening in the future has button to box with membership info


OptimusCrime:
> Påmeldingen er ikke åpnet enda. 
> Mer

Clicking Mer opens dialog with membership information, which makes no sense in this contex.

OptimusCrime:
We could also enhance the template to use the `status_code` instead of looking for variables everywhere. This requires fix for #1379.

https://github.com/dotKom/onlineweb4/blob/develop/templates/events/details.html#L176
https://github.com/dotKom/onlineweb4/blob/develop/templates/events/details.html#L192

sklirg:
Events with closed attendance also does this: https://online.ntnu.no/events/84/hovedekskursjon-2015/

PR #1385, opened 2016-01-20
Fix 'Mer' button appearing when it should not

OptimusCrime:
Fixes issues #1379 and #1378 :smile:


hernil:
Did you check that you are consistent with the status codes in the rest of the events app?

OptimusCrime:
@hernil These codes are not used anywhere else. There is no mention of code 501 anywhere else.

PR #1387, opened 2016-01-20
Resourcecenter refresh

OptimusCrime:
Generelle TODOs:

- [x] Tekst
- [x] Endre rekkefølge på bokser

Bilder
- [x] Online Notifififififfiiyer
- [x] Slack 
- [x] Infosider
- [x] Spillservere
- [x] Kode
- [x] Studentresourouces
- [x] Mailinglister


OptimusCrime:
Not the best images, but it's pretty much what I can do. Feel free to swap or add your own :smile_cat:

hernil:
The images never scale above 201x201px as far as I could see. Thus there is no real point in having the images this large. It just ads to loading times. However I agree it was about time to change them! Some remarks though: 

* Move mailinglister image to spillservere 
* Mailinglister could just have some email icon
* Infosider could just be a thumb of "Gløshaugenkartet"
* Kode would probably look better with coloured code on black background

OptimusCrime:
The biggest images are visible right after bootstrap breaks to mobile view. The image size is 707 pixels at the point. I'll fix the other things.

hernil:
You're right. Carry on then! :-)

OptimusCrime:
How about know @hernil

sklirg:
If you merge/rebase with develop the flake8 errors should pass.

PR #1401, opened 2016-01-29
Revert change when refactoring

hernil:


apps/events/utils.py
```diff
@@ -33,8 +33,8 @@ def get_types_allowed(user):
 
     groups = user.groups.all()
 
-    if reduce(lambda r, g: g.name in ['Hovedstyret', 'dotKom'] or r, groups, False):
-        return [t[0] for t in TYPE_CHOICES]
+    if 'Hovedstyret' or 'dotKom' in groups:
```
sklirg: Matching a string to a Group object? 
``` 
>>> s = User.objects.get(username='sklirg')
>>> 'dotKom' in s.groups.all()
False
```

hernil: Your obviously right. The if just returns True no matter what (what?. Fixing now.

sklirg: Yes :-)

```
>>> 'Hovedstyret' or 'dotKom' in groups
'Hovedstyret'
>>> bool('Hovedstyret')
True
```

PR #1402, opened 2016-01-30
Fix send email to participants

sklirg:


sklirg:
Also: Does not catch all exceptions, only ImproperlyConfigured - which is the relevant one for sending emails.

Issue #976, opened 2014-10-08
Separate settings files


melwil:
We have a few apps that are in separate repositories from ow4, and might require more than a few settings. Would it be cleaner to have these as separate settings files?

My motivation for suggesting this is that base.py gets quite cluttered after a while, and should perhaps include most settings that are connected to django directly. 3rd party apps that I consider taking out settings for in the first place are the ones that are their own systems, like feedme, wiki and redwine.

There are two ways I think this could be done:
- We extract all settings for these systems into their own files, which means having feedme.py, wiki.py and such in the settings folder. This could itself become quite cluttered, and might not be the best solution.
- Second suggestion is just to have a "addons.py" or something like that, and put settings for larger systems in there. This includes adding to installed apps, all SETTINGS_FIELDS and so on.

base.py is on 333 lines right now, and we could easily take out around 50 of these into its own file, perhaps making this worth it in the long run. Any thoughts?

sklirg:
I'm pretty sure we had a chat about this recently, and while `base.py` gets bigger and bigger, adding new empty lines of code to `example-local.py` is a bad way for developers to receive updates. (Who ever overwrites their `local.py`?

I suggest splitting settings files into apps, or at least separate KEYS out of `local.py`, into something like `keys.py`. This file can contain all keys required, e.g. we recently added RECAPTCHA keys, but no one probably noticed (I didn't..) because they are in a file you copy once when you set up the project and then customize to your needs (for logging etc.), and should never have to update again.

melwil:
How about defaults.py for all the settings you would need in dev? And then:

```
if debug: 
    import * from defaults.py
```

The reason we have local.py is not for development, but rather for the production environment, where we want hidden keys and such that are outside the repo, but still in the structure of the code. IF you copy example-local.py when you set up the project, that creates this dummy file, but any settings the project needs would actually be somewhere else.

local.py could still contain stuff like dev emails, dev db, logger settings and so on, but defaults.py would contain any value the project needs to run smoothly beyond the obvious.

sklirg:
Yup! Sounds good to me.

:+1:

sklirg:
I'm working on a project where we tried to separate settings files. A problem emerged and there's no nice solution. How should you proceed if you want to use a setting in a settings file, if it's overridden in local.py?

[...]

PR #1403, opened 2016-01-30
Don't require settings.local

### LINKED ISSUE #1392: Running tox tests locally overwrites local.py
https://github.com/dotKom/onlineweb4/blob/8bd054d3dd44586757692680dae7de327a81ddcc/tox.ini#L25

I'm using postgres as my database backend, and whenever I run tox it overwrites that config in my local.py.

sklirg:
Resolves #1392 

Make all settings have a default value, so we don't require `settings.local` for this project.

This would allow us to run tests without requiring us to copy `example-local.py` to `local.py`. It also makes it possible to run the project without customizing `local.py`. This also means that the test script could be the same at the CI and developer level, since we don't need to copy `example-local.py` to `local.py` for testing anymore.

What this means is that for the project to work correctly, ALL SETTINGS are required in `settings.base`, and the settings you want overridden should be placed in `settings.local` (or as an environment variable). 

### Why?
1) When 30d11f07106db584acd6b42313c01830e3996508 was introduced, every single developer in the project had to open `settings.example-local`, copy four lines regarding some new module into `settings.local` for the project to even run. If we make sure that all settings have a default setting in `settings.base`, we don't need to configure `settings.local` except for when we want to do so.
2) When running tox locally one would overwrite `local.py` with a fresh `example-local.py`, which removes custom `local.py` settings like logging, databases and Stripe / ReCaptcha keys.

### DIFF
--- a/apps/offline/models.py
+++ b/apps/offline/models.py
@@ -7,7 +7,7 @@ from django.db import models
 from django.db.models.signals import post_save
 from django.utils.translation import ugettext_lazy as _
 from filebrowser.fields import FileBrowseField
-from onlineweb4.settings.local import MEDIA_ROOT
+from onlineweb4.settings import MEDIA_ROOT
 from chunks.models import Chunk
 
 import reversion
--- a/onlineweb4/settings/base.py
+++ b/onlineweb4/settings/base.py
@@ -1,6 +1,6 @@
 # -*- coding: utf8 -*-
+import logging
 import os
-import sys
 
 import wiki
 from django.contrib.messages import constants as messages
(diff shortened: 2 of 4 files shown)


hernil:
We've talked about unclogging the base settings file and this doesn't exactly help. We should probably see this in relation to #976

sklirg:
It's not supposed to help that issue. Anyways, see my input on #976 regarding your comment.

If you don't like this solution, what do you think of the case I described under "Why - 1", and how should we solve it? Or is it not an issue?

sklirg:
I'll close this in favour of the ongoing discussion in #976.

PR #1407, opened 2016-02-01
Temporary fix to get Gallery images to fit Nibble

Torrib:
Kanskje min mest sexy pull request.


apps/gallery/util.py
```diff
@@ -153,6 +156,9 @@ def create_thumbnail(source_image_path, destination_thumbnail_path, size):
 
 def resize_image(source_image_path, destination_thumbnail_path, size):
     status = _open_image_file(source_image_path)
+
+    print status
```
OptimusCrime: sweet debug!

OptimusCrime:
Does this in any way fuck up images elsewhere?

Torrib:
It wont change the existing images but it will prevent new images in 16/9 from being uploaded.
This is only a temporary fix while we add the images needed for the inventory then it should be reverted back while we work on a more permanent solution.

apps/gallery/util.py
```diff
@@ -248,7 +252,7 @@ def crop_image(source_image_path, destination_path, crop_data):
 
     # Check all the dimension and size things \o/
     # Give an epsilon of 0.01 because calculations.
-    if crop_width / crop_height < float(16)/float(9) - 0.01 or crop_width / crop_height > float(16)/float(9) + 0.01:
+    if crop_width / crop_height < float(ratio_x)/float(ratio_y) - 0.01 or crop_width / crop_height > float(ratio_x)/float(ratio_y) + 0.01:
```
hernil: Line length

OptimusCrime:
Ah, damn son, that is indeed dirty

sklirg:
Let's wait for Frigg build status.

hernil:
So yeah, I just got cancer. It's temporarily deployed. Will revert asap.

myth:
So yea, how bad was it? Do we have to call an ambulance?

hernil:
It's done, and we do not want this in develop.

PR #1416, opened 2016-02-10
Refactored stripe keys to use dicts

### LINKED ISSUE #1288: Add stripe key for trikom
Trikom needs its own stripe key as demonstrated today with the Star Wars event. I had to use arrkoms key today as I got this on short notice and prokoms key was still in test mode (otherwise we could have switched and added them again later).

### LINKED ISSUE #1310: Change stripe key data structure to dict
Having more than one set of stripe keys means clutter if they are stored in a list. A much safer way is to use a dict for that kind of storage. Not only is it safer, but also a lot more understandable in code.

```python
key = STRIPE_KEYS[0] vs STRIPE_KEYS['arrkom']
```

Torrib:
Fixes #1288, fixes #1310

### DIFF
--- a/apps/payment/migrations/0013_auto_20160210_2010.py
+++ b/apps/payment/migrations/0013_auto_20160210_2010.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('payment', '0012_auto_20160128_1719'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='payment',
+            name='stripe_key_index',
+        ),
+        migrations.AddField(
+            model_name='payment',
+            name='stripe_key',
+            field=models.CharField(default=b'arrkom', max_length=10, verbose_name='stripe key', choices=[(b'arrkom', b'pk_test_replace_this'), (b'prokom', b'pk_test_Ur8B7E5uvheMlpOUVu9SGGbn'), (b'trikom', b'pk_test_bziEErlw9KmTHcNeC6cBXHhw')]),
+        ),
+    ]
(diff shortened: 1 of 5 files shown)


apps/payment/models.py
```diff
@@ -27,16 +27,16 @@ class Payment(models.Model):
         (3, _(u'Utsettelse')),
     )
 
-    STRIPE_KEY_CHOICES = (
-        (0, "Arrkom"),
-        (1, "Prokom"),
-        (2, "Trikom"),
-    )
+    #creates tuples used as key choices.
+    STRIPE_KEY_CHOICES = zip(settings.STRIPE_PUBLIC_KEYS.keys(),settings.STRIPE_PUBLIC_KEYS.keys())
```
hernil: Where does the keys() method come from?

Torrib: Buildt in dictionary functionality

apps/payment/migrations/0013_auto_20160210_2010.py
```diff
@@ -0,0 +1,23 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name='payment',
+            name='stripe_key_index',
+        ),
+        migrations.AddField(
+            model_name='payment',
+            name='stripe_key',
+            field=models.CharField(default=b'arrkom', max_length=10, verbose_name='stripe key', choices=[(b'arrkom', b'pk_test_replace_this'), (b'prokom', b'pk_test_Ur8B7E5uvheMlpOUVu9SGGbn'), (b'trikom', b'pk_test_bziEErlw9KmTHcNeC6cBXHhw')]),
```
sklirg: Keys to the public?

onlineweb4/settings/example-local.py
```diff
@@ -146,17 +146,17 @@
...
+STRIPE_PUBLIC_KEYS = {
+    "arrkom": "pk_test_replace_this",
+    "prokom": "pk_test_replace_this",
+    "trikom": "pk_test_replace_this"
+}
+
+STRIPE_PRIVATE_KEYS = {
+    "arrkom": "sk_test_replace_this",
+    "prokom": "pk_test_replace_this",
+    "trikom": "pk_test_replace_this"
```
sklirg: One is prefixed sk and the others are prefixed pk? Be consistent please, so people who want to debug this uses the correct key.

sklirg:
Looks good to me. :+1:

PR #1421, opened 2016-02-17
Fix displaying extras in dashboard events

cXhristian:
This actually just sends the user directly to the attendees page until we add more tabs. 
The related views needs to be rewritten to proper support tab switching with context like extras.


OptimusCrime:
:+1:

PR #1426, opened 2016-02-24
New favicons

junglebooboo:
Added code for each logo size
Added new favicons + logos


sklirg:
Looks like you're using wrong line-endings. Check https://github.com/dotKom/onlineweb4#installation on how to use the correct one (This is why it looks like you remove all of base.html and frontpage.html, and add it again. This makes it really really hard to read the diff for those files).

junglebooboo:
Fixed! I had to change line separators to CRLF in PyCharm

PR #1428, opened 2016-02-24
Profile-privacy-styling

Aborysa:
Changed the look of the buttons in Profile, "Personvern".

### DIFF
--- a/apps/api/utils.py
+++ b/apps/api/utils.py
@@ -1,0 +1,11 @@
+from rest_framework.response import Response
+from rest_framework.routers import DefaultRouter, SimpleRouter
+
+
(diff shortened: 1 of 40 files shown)


sklirg:
Why are there migrations in here? I guess they come from the "merge branch develop"-commit, which shouldn't be in such a fresh branch. Either close this PR and open a fresh one with the changes (freshly branched out from develop) or rewrite the history so you don't get that merge-commit (first one is the easiest).

Also, it is nice to name the branch something which connects it to the issue at hand.

Aborysa:
@sklirg Yes, was an old test branch I re-used, will try to create a new one.

Issue #1363, opened 2015-11-19
500 Internal Server Error template


sklirg:
Show something more than

# 500

If a user triggers a 500 INTERNAL SERVER ERROR.

Could provide a form for what the user did to come to this place, which sends an email to dotkom.

PR #1439, opened 2016-03-02
created simple 500 page

nornes:
A very simple page, doesn't scale and doesn't inherit anything from other files (except an image).

### DIFF
--- a/templates/500.html
+++ b/templates/500.html
@@ -1,0 +1,68 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+    <title>Internal server error - Online</title>
+    <meta name="description" content="Online er linjeforeningen for informatikkstudentene ved NTNU. Våre studenter går Bachelor- eller Mastergradstudium i informatikk."> 
+    <style type="text/css">
+
+
+@font-face {
+  font-family: 'Open Sans', arial, 'sans-serif';
+  font-style: normal;
+  font-weight: 300;
+  src: url('/static/fonts/open-sans-300-webfont.eot');
+  src: url('/static/fonts/open-sans-300-webfont.eot?#iefix') format('embedded-opentype'), url('/static/fonts/open-sans-300-webfont.woff') format('woff'), url('/static/fonts/open-sans-300-webfont.ttf') format('truetype'), url('/static/fonts/open-sans-300-webfont.svg#open_sanslight') format('svg');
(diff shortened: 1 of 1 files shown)


nornes:
Fixes #1363

sklirg:
I think you safely can extend base.html, and add a form to the page. 

While it's nice that the page is as simple as possible, we also want users to feel inclined to actually provide feedback, which is easier if we provide them all the means to do it, rather than them having to compose and send an email. Suggestions for the form is a required text area where the user can describe what went wrong, and an optional (or required) name and/or email-field (for future contact).

My few cents :+1:

cXhristian:
500 templates are rendered without any context which might cause problems?
It would also be a bit silly if one of the template tags caused the server error which again fucked up the error page.
It might work just fine though.

sklirg:
> 500 templates are rendered without any context which might cause problems?

Rendered, yes. Whatever you do on it, however, contains context, such as a form. Would require a separate view to handle form submit, however.

> It would also be a bit silly if one of the template tags caused the server error which again fucked up the error page.

I agree. However, a white page is ugly. I'd like to have something similar to 404 and 403 pages, with a form instead of a nyan cat. Those are my thoughts though, if everyone thinks mail will suffice, I guess that's fine. :stuck_out_tongue:

cXhristian:
>Rendered, yes. Whatever you do on it, however, contains context, such as a form. Would require a separate view to handle form submit, however.

I meant context variables that base.html uses. But looking more closely into it STATIC_URL will still be included and if checks like user.is_authenticated should still work fine. 

I agree that a form would be nice.

sklirg:
Okay. Create a super simple version of base.html and add a form to it? 😊

nornes:
I found a lot of ambiguous information when googling this, so I just ended up doing it the safe and simple way. The problem seems to be that django doesn't render static files in error views, but I think I've figured out the solution now, so I'll try making another one this weekend if I get the time. 😊

myth:
> The problem seems to be that django doesn't render static files in error views

Where did you read this? Our 404 and 403 templates render just fine (stylesheets are staticfiles too).
I use custom error templates on my own site, and just extend `base` as commented previously, on 400, 403 and 500 templates.

Additionally, if we want to be able to include custom context, and post the report form directly to the view, one can replace the existing default error view for django, by simply setting `handler500` to a custom view (i.e `'onlineweb4.views.server_error'`) or something of the sort.

A possible example:

```
{% extend 'base.html' %}
{% block title %}Ooops{% endblock %}
{% block content %}{{ error_form }}{% endblock %}
```

Create a `views.py` in the `onlineweb4` folder for generic, site-wide views.
Create a `server_error` view with some custom error form imported and added to the context.

And in `onlineweb4.urls`:

```
from onlineweb4.views import server_error

handler500 = server_error
```

PR #1440, opened 2016-03-02
Redesigned the chevrons for the Offline carousel

### LINKED ISSUE #1389: Reduce size of carousel arrows in the `Offline` section
Bad ratio between button (the grey background) and the white arrow in the `Offline` section on the front page

### LINKED ISSUE #1424: Focus stays on chevrons on Offline section
If you click on the carousel chevrons in the "Offline" section of the frontpage, the the focus changes their color to a darker gray/black. After the transition is complete, the focus/color stays. Is this intended/wildly unimportant, or a mere oversight?

sardred:
Fixes #1389 by redesigning the chevrons completely. The redesign falls more in line with the design of the site, as the previous chevrons utilized an aesthetic we don't use anywhere on our sight. I went with the blue over orange, because orange is typically a color that is un-interactable on the site, while links and interactable objects are typically blue.

Also fixes #1424 by simply removing the focus styling.


cXhristian:
I'm not sure if it's a good idea to edit files/static/less/bootstrap/carousel.less. 
It could easily get replaced when upgrading to a newer bootstrap version.
Also edit bootstrap-variables.less instead of variables.less for the same reason.

Do you think it is doable to fix this without editing carousel.less directly?

sklirg:
Also, seems like you branched out from a branch that's ahead of develop. It will probably get merged soon anyways, but keep that in mind for next time (this simple graphical fix would not be able to merge simply because of database changes not relevant to the raised issue).

sardred:
My bad with the branching there. You learn from your mistakes, I suppose. I'm not sure what the bootstrap directory means, but I couldn't find the styling anywhere else in the .less files (which I'm not well versed with either) so I don't know if it could be solved in another fashion. I'm also not sure what use bootstrap-variables.less would have in this scenario. 

I won't lie, I have a hard time imagining how to remove a gradient which comes from a particular file, if I'm not allowed to edit that particular file.

sardred:
I've looked into it a bit now, and by editing the offlinewidget.less file I could overwrite all the styling from the bootstrap files that I want to get rid of, but it's a sort of roundabout way of doing it and it seems like clutter.

cXhristian:
Yeah, I get what you mean. It's just so much easier to upgrade bootstrap if you can replace all the files in the bootstrap folder without worrying about custom code we have added. 

The bootstrap folder is basically the source code download from http://getbootstrap.com/getting-started/#download. 

I don't have a good solution other than overwriting all the styles unfortunately (which isn't very efficient either).

OptimusCrime:
Please overwrite in offlinewidget.less :)

sardred:
After mucking about with git for a while, I've finally redone it with only overwriting in offlinewidget.less

OptimusCrime:
Looks good :+1:

PR #1443, opened 2016-03-02
Remove broken link in adminpanel title

nixolas1:
Yes.


cXhristian:
Would it be possible to still have a link back to /?

hernil:
Isn't this fixed as of now? https://online.ntnu.no/admin/

cXhristian:
It's broken on dev.

sklirg:
Grappelli probably decided it was smart to strip HTML tags. Which is smart.
So the link-hack doesn't work. However, there's a link that's called "View
webpage", which does the same thing.

cXhristian:
Oh, yeah. That's true. Merging this then.

PR #1444, opened 2016-03-02
Fixed #1441

### LINKED ISSUE #1441: Popover overlaps menu
The .popover class in the bootstrap has a z-index of 1060 while the #mainnav has a z-index of 10.

Since the modals also use a z-index at 1050 I assume the .popover class need to be set to lower than 10 rather than changing the #mainnav to more than 1060 (then the menu also overlaps the modals).

The issue was observed when editing profile at /profile/edit/ and selecting the picture's info-mark - then scrolling down so the fixed navbar overlaps the popover from the info-mark.

niklasmh:
Set z-index on .popover and .dropdown-menu to 9 since they both caused the same issue. Did not register any following issues after the change.

### DIFF
--- a/files/static/less/core.less
+++ b/files/static/less/core.less
@@ -187,3 +187,7 @@ ul li {
 .navbar li {
     padding: 0;
 }
+
+.popover, .dropdown-menu {
+    z-index: 9;
+}


sklirg:
You merged your own PR? :scream:

niklasmh:
Oops.. I know I was a little fast on the merge-button (long time ago, but now I'm back in the flow).

sklirg:
😀😎

PR #1445, opened 2016-03-02
Fixed #1386 one class with datetime, one with date only and one with just t…

### LINKED ISSUE #1386: Extend dashboard DTP with time of day
Currently only the date portion of the datetimepicker is active.

Reconfigure `dashboard.js` to include the time of day as well.

niklasmh:
…ime. Elements affected by the .dpt-class will change to datetime unless only date is specified with the .dp-class, or only time is specified with the .tp-class (every element is now .dtp which is a full datetime).


sklirg:
> (every element is now .dtp which is a full datetime)

Have you made sure alle fields that depends on the current implementation of datetimepicker correctly works with this new version?

niklasmh:
The parser should parse the format YYYY-MM-DD HH:mm:ss as it is only timestamps, but I'll look into it to be sure.

niklasmh:
It is tested and should work.

sklirg:
There shouldn't be a timestamp in `/dashboard/posters/add/`, only date.

screenshot 2016-03-05 17 43 03

OptimusCrime:
Why not do this the other way around, so the existing class is only date, and new classes are datetime and time? This will fuck up every place that is using only date today, which I guess are a few.

myth:
> Why not do this the other way around, so the existing class is only date, and new classes are
> datetime and time? This will fuck up every place that is using only date today, which I guess are a few.

1. Because `.dtp` is datetimepicker, `.dp` is datepicker and `.tp` is timepicker.
2. No, there are actually very few places that incorporate the datepicker, so removing one letter from a couple of templates is a very small job.
3. Part 2 of this is to add django field widgets with proper template code for use in our forms

OptimusCrime:
Aren't these classes custom? Meaning we can call it whatever we like? I understand the abbreviation, but it seems like more trouble doing it this way.

However, if there is only a few places where this is actually used, then go ahead, but remember to update every single place it is in use.

myth:
Yes these are custom, no there is no trouble. It was a mispronounciation in the first place. This PR fixes that bug/mispronounciation. You don't leave stuff like these being semantically wrong just because it's "less trouble".

PR #1446, opened 2016-03-02
Made event extras visible until registration_deadline

nixolas1:
Instead of unattend_deadline. The issue was that people could attend an event and not be able to choose a food thing.


sklirg:
What happens if `unattend_deadline` is after `registration_deadline`?

sklirg:
Bump @nixolas1 ?

nixolas1:
Could not test (broken python installation). Needs good testing before merge.

PR #1448, opened 2016-03-02
Closer eventtext

niklasmh:
Fixed the issue for both desktop and mobile with float: right; on the event-status. Everything now looks like before with just less space as requested.

PR #1450, opened 2016-03-02
Fixed space between event text and event information

niklasmh:
Fixed the issue for both desktop and mobile with float: right; on the event-status. Everything now looks like before just with less space as requested. (Dubplicate of [#1448](https://github.com/dotKom/onlineweb4/pull/1448))

### DIFF
--- a/files/static/less/event_details.less
+++ b/files/static/less/event_details.less
@@ -139,9 +139,13 @@ section#event-details {
             color: white;
         }
     }
-    .event-status-container {
-        @media (min-width: 992px) {
+    @media (min-width: 992px) {
+        .event-status-container {
             float: right;
+        }
+        .event-status-companies-container {
+            float: right;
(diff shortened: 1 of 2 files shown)


hernil:
Print screens in the PR are great for illustrating design changes like these btw. :smile:

niklasmh:
Width more than or equal 992px:
image

Width less than 992px:
image

niklasmh:
Also one last situation; When the event status (orange box) makes the right side taller than the event text on the left:
image

OptimusCrime:
Hmmm, that is not good. Perhaps make the final box there float to the left for wide devices or something?

niklasmh:
It is a really rare situation, but I agree, it should not happen.

niklasmh:
Now it is fixed and looks like this in the same situation:
image

PR #1452, opened 2016-03-03
EksKom can mail participants for "ekskursjoner".

sklirg:
Allows eksKom members to send mail to attendees for excursion events. This is in tag 4.17.9 already, but required in develop for consistency.


sklirg:
Is already in prod, so no reason to let this PR be open. KTHX

PR #1460, opened 2016-03-08
Fixed #1453

### LINKED ISSUE #1453: Logo and links disappearing when site is resized
The logo and links in the header will disappear when window width is reduced, menu is opened and closed, and width is put back to fullscreen. 

Steps to reproduce: 
Reduce window width to less than 770px
Open hamburger menu
Close hamburger menu
Resize window to more than 770px

niklasmh:
Generated inline-style from jQuery are now plain CSS classes - which is better for future changes. Also replaced some old vanilla JS with build-in jQuery functions - since they are more robust and easier written.

### DIFF
--- a/files/static/js/init.js
+++ b/files/static/js/init.js
@@ -100,16 +100,20 @@ $(function() {
 
     /* nav bar toggle
     ---------------------------------------------------------------------------*/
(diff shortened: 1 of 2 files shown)


hernil:
Seems to work well enough.

PR #1466, opened 2016-03-09
Fixed inventory ordering by id

Torrib:


OptimusCrime:
Why not name? :)

sklirg:
Ordering by PK is cheaper in terms of performance, as name field most likey isn't indexed.

OptimusCrime:
@sklirg Yes, that is true, but the performance impact is so small it is next to nothing considering our tiny table. I was thinking about what made most sense in the actual inventory screen. Even though the products are ordered by ID it makes sense in a "backend" kind of way, but for people browsing the inventory it is more or less the same as a random ordering.

sklirg:
When exposing [id, name, price](https://github.com/dotKom/onlineweb4/blob/e7252f5b47a3c1c4a01aad1cd676cd29003c14b6/apps/shop/serializers.py#L66) etc. it's perfectly reasonable to order the items by id and rather allow the front-end to sort items on whatever it wants.

PR #1467, opened 2016-03-09
Fixed #1462

### LINKED ISSUE #1462: Top bar blocks content on small devices
After changing the z-index of the top menu bar its sub nav bar is stopping clicks from passing through even though it is not visible.

niklasmh:
Made an animation-in-progress class instead of animation-complete which needed initialization. Now it works as first thought.


hernil:
Testing on mobile doesn't seem to work.

niklasmh:
Should work now. Fixed another issue related to z-index on popover and dropdown-menu classes. Popover seems to work fine on every device, but changing the z-index on the drowdown-menu from the bootstrap was not a good idea. Anyway, it is fixed, and please comment if you find some more issues related to the navbar which wasn't there before.

niklasmh:
Removed dat big a** core.css file, so merging should be good now.

PR #1474, opened 2016-03-10
[WIP] Full Gallery front end re-implementation

myth:
This branch will do a complete re-design and implementation of how the Gallery front end works. There are several flaws and errors that need to be fixed, as well as the new version of Cropper.js having a somewhat new API.

Todo:

- [x] Upgrade `Cropper.js` to newest version
- [x] Get rid of old js
- [x] Create a minimalist event system instead of polling
- [x] Redo the Gallery module's handling of pane selection and click events
- [x] Redo how `UnhandledImage`s are displayed in the manage pane
- [x] Implement the cropping view using the new `Cropper.js` dynamic options
- [x] Implement support for the presets introduced in the Gallery backend refactor branch
- [x] Include a microlib for HTML templating (needed for widgets, ajax search etc).
- [x] Re-write the widget code for `ResponsiveImage` form fields
- [x] Add additional cropper controls and view modes as seen on http://cropperjs.com
- [x] Implement additional validation checks and client side feedback
- [x] Make the client fetch crop templates from the server-side configuration
- [ ] Refactor and clean up the views, extract common stuff to mixins and convert to CBVs
- [x] Enable front end crop data submit to back end and test rigorously
- [x] Ensure documentation (docstrings and comments) are up to date and sufficient


OptimusCrime:
underscore.js is a great lib for HTML templating.

myth:
Underscore is massive tho. Was thinking more along the lines of T.js, Transparency, or something similar that just handles templating (like Handlebars), and not everything else aswell.

Was going to use Transparency because the simplicity of the API is just awesome, but it hasn't been maintained for a few months and i'm experiencing a bug.

http://jsperf.com/transparency-vs-handlebars-infinite-list

Performs quite well, so I'm still looking for something similar.

myth:
On second thought, underscore offers quite a few things we need, so we could just go for it regardless.

OptimusCrime:
Yeah, underscore has a lot of functionality that we can use besides the template functionality. I only mentioned it because I've had great experience with it in the past. However, I guess you can find identical functionality in other libraries as well.

Also, it looks like underscorejs i still in active development? https://github.com/jashkenas/underscore/commits/master

myth:
Ah, I just checked their releases tab. I can see there's a lot of underscore vs lodash discussion around, with talks of merging the two aswell. So i guess we can pick and choose then, which is nice.

OptimusCrime:
Sure :)

myth:
Underscore is now included in our codebase.

myth:
I think this branch has enough glue and makeup on now that it can be merged and tested.

PR #1481, opened 2016-03-13
Make it possible to add studmail after deleting it. Fix #1480 

### LINKED ISSUE #1480: Deleted email address can't be added back
Deleting an email address from the user profile makes it impossible to add that same address back. Attempting to add it back gives a message saying a user with the NTNU-username is already registered. 

Problem persists weeks after the address was removed.

nornes:
Not 100% sure if this is the solution, but it seems likely as the filtering of User objects return a list.


myth:
That should do it, yes.

PR #1482, opened 2016-03-13
Group sync

hernil:
Disconnects signal hooks as to avoid infinite recursive calls when adding or removing users to syncable groups.

### DIFF
--- a/apps/authentication/signals.py
+++ b/apps/authentication/signals.py
@@ -1,4 +1,7 @@
 # -*- coding: utf-8 -*-
+import logging
+import uuid
+
 from apps.authentication.tasks import SynchronizeGroups
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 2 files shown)


myth:
Seems like you managed to figure out how dynamic dispatching works! :)
Other than some inconsistencies regarding log format strings, it looks good.

Which variations of group operations have you tested? Eager to test this on dev first.

hernil:
It should cover adding/removing users to and from groups, adding/removing groups to and from users, as well as be triggered when saving a group object (debatable whether this is necessary, but was already there from @sklirg and did not add any complexity). 

This is not triggered when editing the group-sync settings, however it will run just fine next time someone changes any one of the objects mentioned above.

hernil:
I do have a weird problem with log entries being duplicated, but I suspect the problem to be local.

PR #1487, opened 2016-03-16
[WIP] Internal access backend

krekle:
This branch aims towards a oneway sync between ow4 and LDAP. Users can set their LDAP password in their personal profile. The LDAP database is used for file access and SSH stuff.

List of things to do:
- [x] Form for setting password
- [x] Make sure the user fill in their correct password before they can set their LDAP password
- [x] Fix pep8 errors in Profile
- [x] Make successful connection with LDAP (this will most likely never happen)
- [x] Make everything work  
- [x] Fix insert new user
- [ ] Insert samba schema in LDAP
- [ ] Hash NTLMv2 password and store in LDAP
- [ ] Setup samba server and configure to use LDAP as auth backend
- [ ] Test samba shit

This might take a while.


hernil:
Been looking at this with @iverasp tonight. Updating users in ldap should be good, but some work remains on creating new users.

iverasp:
Ser denne ble satt rett i prod. Det er jo kult. Før den er heelt ferdig må følgende punkter fullføres:

- [ ] Primary group based on group priorities (prokom before dotkom, etc). This does not currently work correctly.
- [ ] Create new LDAP groups if they are created in OW4
- [ ] Let user choose shell
- [ ] Password complexity requirement on backend
- [ ] Password complexity requirement on client
- [ ] Test that everything works

PR #1488, opened 2016-03-19
Update vagrantbootstrap.sh with python3 support

cXhristian:
I didn't actually test this with vagrant, but I don't see why it shouldn't work.


sklirg:
Why not use `python3` straight from path?

cXhristian:
I remember having trouble with that on some systems. Or it might have been an older version of virtualenvwrapper. I tested with just python3 locally and it worked now though.

PR #1489, opened 2016-03-23
Genfors blank votes

MrHanSan:
Updated the genfors-app, so it works again. Changed how blank votes affect the result, and how blank votes are represented on the page. Also changed vote verification so it disappears if option is changed.


files/static/js/genfors.js
```diff
@@ -123,40 +129,55 @@ Genfors = (function () {
...
+                            else{
+                                percent = Genfors.get_percent(value, votes_for_alternative);
                             }
-                            else if (key == 'Nei') {
-                                type = 'danger';
+                            html += '<strong>' + key + '</strong>: ' + value + ' stemme' + ((value>1)?'r':'');
+                            type = "";
+                            if (key == 'Blankt') { 
+                                type = 'warning';
+                                html += '<p /><div class="progress"><div class="progress-bar progress-bar-' + type + '" role="progressbar" aria-valuenow="' + percent +
```
sklirg: Empty `<p>`-tags should not be used. If you're looking for blank lines, use CSS (margin) or `<br />`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p#Notes

sklirg:
Also, have a look at the failing Flake8 tests :)

hernil:
Calculation of votes does not yet account for (/subtract) blank votes.

PR #1492, opened 2016-03-30
Password authentication

### LINKED ISSUE #1436: More informative set_password error messages and some authentication fixes
In [`apps.authentication.views.set_password`](https://github.com/dotKom/onlineweb4/blob/v4.17.8/apps/authentication/views.py#L216) there are some edge cases not being handled gracefully, resulting in a HTTP 500 INTERNAL SERVER ERROR if triggered. 
- [x] `if rt.is_valid:` does not have an else, and there's no return statement - which means if the token is invalid for whatever reason (expired or a "fake"), an exception is thrown due to a view not returning a `HttpResponse` object. Fix by returning something, e.g. a `request.message` and a redirect to root (`/`).
- [x] Make sure any other cases returns to root with a message, simply remove L 250 `else` and unindent, adding relevant `request.messages` where appropriate.
- [x] Add minimum length password validators according to [Django 1.9 password validators](https://docs.djangoproject.com/en/1.9/topics/auth/passwords/#enabling-password-validation).
- [ ] _up for discussion_ Update RegisterToken model to have `RegisterToken.token` field as unique. If that's implemented, we can remove `if token.count() == 1` and replace it with a `get_object_or_404`, reducing code complexity, branching and readability.

nornes:
fix #1436 
Also adds more logging to authentication

### DIFF
--- a/apps/authentication/views.py
+++ b/apps/authentication/views.py
@@ -2,6 +2,7 @@
 
 import re
 import uuid
+import logging
 from smtplib import SMTPException
 
 from django.conf import settings
@@ -85,7 +86,7 @@ def register(request):
 
                 # Create the registration token
                 token = uuid.uuid4().hex
-                try: 
+                try:
                     rt = RegisterToken(user=user, email=email.email, token=token)
                     rt.save()
                     log.info('Successfully registered token for %s' % request.user)
@@ -187,13 +188,12 @@ def recover(request):
 
                 # Create the registration token
                 token = uuid.uuid4().hex
-                try: 
+                try:
                     rt = RegisterToken(user=email.user, email=email.email, token=token)
                     rt.save()
                     log.info('Successfully registered token for %s' % request.user)
                 except:
                     log.error('Failed to register token for %s' % request.user)
-                
                 email_message = _("""
 Vi har mottat forespørsel om å gjenopprette passordet for kontoen bundet til %s.
 Dersom du ikke har bedt om denne handlingen ber vi deg se bort fra denne eposten.
@@ -227,7 +227,7 @@ def set_password(request, token=None):
     if request.user.is_authenticated():
         return HttpResponseRedirect('/')
     else:
-        try: 
+        try:
             rt = RegisterToken.objects.filter(token=token)[0]
             if rt.is_valid:
                 if request.method == 'POST':
@@ -259,7 +259,7 @@ def set_password(request, token=None):
                     _('Lenken er ugyldig. Vennligst bruk gjenoppretning av passord for å få tilsendt en ny lenke.')
                 )
                 return HttpResponseRedirect('/')
-        except: 
+        except:
             log.debug('User %s failed to recover password with token %s' % (request.user, rt))
             messages.error(
                 request, 'Noe gikk galt med gjenoppretning av passord. Vennligst prøv igjen.')
--- a/apps/authentication/views.py
+++ b/apps/authentication/views.py
@@ -1,8 +1,7 @@
 # -*- coding: utf-8 -*-
-
+import logging
 import re
 import uuid
-import logging
 from smtplib import SMTPException
 
 from django.conf import settings
--- a/apps/authentication/views.py
+++ b/apps/authentication/views.py
@@ -47,6 +47,8 @@ def logout(request):
 
 @sensitive_post_parameters()
 def register(request):
+    log = logging.getLogger(__name__)
+
     if request.user.is_authenticated():
         messages.error(request, _('Registrering av ny konto krever at du er logget ut.'))
         return HttpResponseRedirect('/')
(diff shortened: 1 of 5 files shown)


sklirg:
This branch is missing the required migration for change to RegisterToken-model. Check #1494.

apps/feedback/models.py
```diff
@@ -171,9 +171,9 @@ def save(self, *args, **kwargs):
             try:
                 rt = RegisterToken(fbr=self, token=token)
                 rt.save()
-                log.info('Successfully registered token for %s' % rt.user)
+                log.info('Successfully registered token for fbr %s with token %s' % (self, rt))
```
sklirg: We don't need to shorten words to weird abbreviations, so next time, please use the full name. :)

apps/authentication/views.py
```diff
@@ -83,9 +85,12 @@ def register(request):
...
                 token = uuid.uuid4().hex
-                rt = RegisterToken(user=user, email=email.email, token=token)
-                rt.save()
-
+                try:
+                    rt = RegisterToken(user=user, email=email.email, token=token)
+                    rt.save()
+                    log.info('Successfully registered token for %s' % request.user)
+                except:
+                    log.error('Failed to register token for %s' % request.user)
```
sklirg: Why would this fail?

apps/authentication/views.py
```diff
@@ -208,13 +222,12 @@ def recover(request):
...
+    log = logging.getLogger(__name__)
     if request.user.is_authenticated():
         return HttpResponseRedirect('/')
     else:
-        tokens = RegisterToken.objects.filter(token=token)
-
-        if tokens.count() == 1:
-            rt = tokens[0]
+        try:
+            rt = RegisterToken.objects.filter(token=token)[0]
```
sklirg: Much easier to use `RegisterToken.objects.get(token=token)` :-)

onlineweb4/settings/base.py
```diff
@@ -144,16 +144,6 @@
...
-#Django 1.9 passwordvalidation
-AUTH_PASSWORD_VALIDATORS = [
-    {
-        'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
-        'OPTIONS': {
-            'min_length': 6,
-        }
-    },
-]
-
```
sklirg: Why remove this again?

sklirg:
Also, according to https://github.com/dotKom/onlineweb4/pull/1492/commits/acf5a4db7e51a8728af5afd78d4f35da3e64b232#diff-70bc887b030b5f962808f915c24e3e68L147, Password validation was removed again.

onlineweb4/settings/base.py
nornes: Didn't mean to do that, I reverted the entire base file to the develop version when I noticed my changes from working on the 500-page was still on mye computer, and I forgot I'd changed anything. I'll put it back in when I get back on the right computer ;)

apps/authentication/views.py
nornes: I believe @myth was the one to puth the try-except block in there. After making token a unique field in the model it would probably fail if the token has been used before i guess(?).

myth: I did? Can't remember, anyway, the except should not be without an explicit exception. E.g `except IntegrityError` or similar. And it can certainly fail, as all things can. But with the new 500 temlate, it should log interesting information and then re-raise the exception to render the template.

nornes: Now that you mention it I do remember you also said I should change it to account for IntegrityError, I just forgot to actually do it.

sklirg: This way, if the code fails it'll just keep going. What needs to happen, is to make the program generate a _new_ code, so it actually makes sense to continue. 

We need the program to create a valid code.

This functionality could override the `create()`-function of the object, or create a custom function which we use instead of the default `create()`.

nornes: Like creating new tokens until one works you mean? I might be asking stupid questions here, but wouldn't putting that in the `create()`-function make the token different from the one in the program that creates a RegisterToken object? Not sure if that would matter at all though. Also, if the token is changed upon creating the object anyways, would it even need to be passed in as an argument. 

Having the creation of the token in a loop running until it passes seems easier, but that might be a bad way of doing it as it would have to do that every time a RegisterToken is created.

sklirg: Yes. No, not at all. This is a bit of a weird case. 

By overriding the `create()`-method, we'll override the `create()`-method for _all_ RegisterToken objects. We'll also make sure that it behaves just like it does today, but move the generation of a unique token to the `create()` method instead of manually generating it first, and then adding it to the model.

Yes, we have to keep generating codes until we find a string that works. However, the [probability of collision](https://en.wikipedia.org/wiki/Universally_unique_identifier#Random_UUID_probability_of_duplicates) is very small, though it exists. This means we most likely won't ever need to re-generate a code, and if we do, we'll most likely not have to do it a lot of times.

Here are some of my thoughts in some kind of code.

```
registertoken = RegisterToken.objects.create(user=user, email=email)

# Django will now create an object for user and email, 
# which will automagically generate a token on the object like so:

>>> registertoken.token 
a494f4101ad54f2bb91b82612bf6320d  # uuid.uuid4().hex
```

This way we don't need to remember to check for integrity errors every time we create a token.

Or, we can do it like this
```
def generate_unique_token():
    token = uuid.uuid4().hex
    if RegisterToken.objects.filter(token=token):
        token = generate_unique_token()
    else:
        return token

...

token = generate_unique_token()
RegisterToken.objects.create(user=user, email=email, token=token)
```
or, as you suggested, a loop:
```
token = uuid.uuid4().hex
while RegisterToken.objects.filter(token=token):
    token = uuid.uuid4().hex

RegisterToken.objects.create(user=user, email=email, token=token)
```

The first way I suggested would implement either of these two checks, or some other, to make sure a unique token is generated. This could be implemented by hooking onto a pre_save signal (an event which is fired _before_ a model is persisted in the database).

In my opinion, the first option is the cleanest. It will place the token validation in one place, and developers won't have to think about verifying if the token already exists, Django will take care of that in the future.

PR #1498, opened 2016-04-04
make the form not send email when no message is provided

### LINKED ISSUE #1496: Don't send additional 500 email if no useful information in it.
After the implementation if #1485 we get an additional email for _all_ 500 errors, even if there is no supplied information. Let's not send an email if there's no useful information in it.

nornes:


onlineweb4/views.py
```diff
@@ -1,21 +1,22 @@
...
 
 def server_error(request):
     log = logging.getLogger(__name__)
 
     if request.method == 'POST':
-        form = ErrorForm(request.POST)
-        message = request.POST.get('reason', 'Ingen forklaring oppgitt.')
-        try: 
+        message = request.POST.get('reason')
+        if message is None:
```
sklirg: It would suffice to do `if message:`, since `if message:` would call `bool()` on `message`, resulting in `bool(message)` => `True` for anything other than `""` (empty string), which would be `False`. 

So if an empty form is submitted, or nothing is submitted, this would be `False`, resulting in a redirect to home.

TL;DR: `if message:` (instead of `if message is None:`) would also catch an empty form.

sklirg:
:+1:

apps/authentication/views.py
```diff
@@ -227,7 +227,7 @@ def set_password(request, token=None):
         return HttpResponseRedirect('/')
     else:
         try:
-            rt = RegisterToken.objects.filter(token=token)[0]
+            rt = RegisterToken.objects.get(token=token)
             if rt.is_valid:
                 if request.method == 'POST':
                     form = ChangePasswordForm(request.POST, auto_id=True)
```
nornes: Didn't mean to add this commit to this branch, not sure how to remove it.

onlineweb4/settings/base.py
```diff
@@ -144,6 +144,16 @@
...
+        'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
+        'OPTIONS': {
+           'min_length': 6,
+        }
+    },
+]
+
 ROOT_URLCONF = 'onlineweb4.urls'
 
 # Python dotted path to the WSGI application used by Django's runserver.
```
nornes: Didn't mean to add this commit to this branch, not sure how to remove it.

nornes:
fix #1496

PR #1513, opened 2016-04-13
Membership expiration

### LINKED ISSUE #1400: Users applying for membership changes with field of study gets expiredate in the past
If you apply for membership today with a Bachelor started in 2011, you will get an expiry date of 2014-09-16.

The given expiration date for a membership should be adjusted upward if the suggested duration of a field of study puts this date in the past.

nornes:
Fix #1400
Sets the expiration date for membership-applications where estimated expiry is in the past to upcoming September.

### DIFF
--- a/apps/approval/views.py
+++ b/apps/approval/views.py
@@ -70,7 +70,7 @@ def create_fos_application(request):
 
 
 def get_expiry_date(started_year, length_of_fos):
-    today = timzone.now().date.today()
+    today = timezone.now().date()
     # Expiry dates should be 15th September, so that we have time to get new lists from NTNU
     new_expiry_date = datetime.date(
         started_year, 9, 16) + datetime.timedelta(days=365*length_of_fos)
--- a/apps/approval/views.py
+++ b/apps/approval/views.py
@@ -6,6 +6,7 @@ from django.contrib import messages
 from django.contrib.auth.decorators import login_required
 from django.http import Http404
 from django.shortcuts import get_object_or_404, redirect
+from django.utils import timezone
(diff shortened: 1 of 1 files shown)


apps/approval/views.py
```diff
@@ -70,6 +68,20 @@ def create_fos_application(request):
     raise Http404
 
 
+def get_expiry_date(started_year, length_of_fos):
+    # Expiry dates should be 15th September, so that we have time to get new lists from NTNU
+    new_expiry_date = datetime.date(
+        started_year, 9, 16) + datetime.timedelta(days=365*length_of_fos)
+    # Expiry dates in the past sets the expiry date to next september
+    if new_expiry_date < datetime.date.today():
```
sklirg: Use `django.utils.timezone` instead of `datetime.datetime`. :-) That takes care of all timezone stuff (and DST/Summer time).

nornes: I can see how that is better on a general basis, but is it really necessary here when it's only dealing with dates? Even if it gets set off by one day due to time-differences that wouldn't be that bad, would it?

sklirg: https://docs.djangoproject.com/en/1.9/topics/i18n/timezones/#naive-and-aware-datetime-objects

This matters mostly for datetime objects, so you're right in the fact that it's less important for date. Anywho, it's better practice to do `today = timzone.now().date.today()` and work with `today` rather than doing `datetime.date.today()` every time you need the variable (the same goes for `datetime.date.today()` for that matter). :-)

So yes, I was a bit too quick to respond in this case.

PR #1517, opened 2016-04-13
Fix extras counting

sklirg:
Also refactors some names in the function so it makes sense.


sklirg:
screenshot 2016-04-13 21 30 20

PR #1530, opened 2016-04-16
Gallery upload size fix

myth:
Corrects an error causing image uploads to fail, due to django using different file handler classes based on file size.


sklirg:
I renamed the context from teamcity from teamcity -> teamcity-pytest which makes it look like it's not done. Both of the "details" links links to the same page.

Issue #1376, opened 2015-12-14
Replace django-chunks


sklirg:
We should replace [django-chunks](https://github.com/clintecker/django-chunks). It's not actively maintained and it fails on Django 1.9 because it uses deprecated functions.

Our uses:
1. Frontpage, infotext regarding each committee.
2. Information texts for Offline

The functionality of 1 is scheduled to be removed in #1209 (frontpage, about all committees).
The functionality of 2 needs to be completely replaced.

As of now this module is hindering a Django version upgrade so we might want to do some temporary fix for this, and implement the change coming in #1209 at a later time. This way, we can upgrade to Django 1.9 asap.

Suggestion for temporary fixes:
- Install https://github.com/funkybob/django-flatblocks and port the data over to that module (or another similar module)
- Move current chunks into the html-templates, effectively hard-coding them into there for the time being (this removes edit-functionality from the administration panel for the time being)
- Flatpage for "about us"

myth:
I believe Django already has built-ins that can be used. The flatpages app is an optional that basically does the same thing.

sklirg:
The problem with flatpages is that it renders a whole page, which would not work for our frontpage  (which has events, articles etc.). 
We could, however, move that content to a separate page (about us) and show it there. That page could then be updated upon merging of a fix for #1209.

myth:
Reuse just the models then, or roll our own. It's not like it's much complexity in chunks.

myth:
Bump, since the complexity is very low, and this is an excellent entry-level project for the new guys to get to know python package development in general and django addons in particular. Since chunks is no longer maintained, we should task ourselves with creating a similar package, and also add some additional features wee see fit.

sklirg:
I forked it and fixed the bugs for the 1.9 PR. Fwiw.

sklirg:
Should we write some specs for this module, so any takers know what's required?

sklirg:
Since this mainly was a blocker for bumping to Django 1.9, I'll close this for now. If we want to, we can modify the "chunks" to basically be the whole about us page, and provide API endpoints for them instead of template tags.

PR #1532, opened 2016-04-19
Add dashboard module for chunks

sklirg:


myth:
Looks good. Missing `DeleteView`? Also, about the directory placement. I havent really thought about it, but for now placing it under dashboard is probably OK. If we figure out some better way in the future it's no biggie to move it.

apps/dashboard/chunks/dashboard/views.py
```diff
@@ -0,0 +1,41 @@
...
+class ChunkCreate(DashboardPermissionMixin, CreateView):
+    model = Chunk
+    fields = ('key', 'description', 'content')
+    permission_required = 'chunks.add_chunk'
+
+    def get_object(self, queryset=None):
+        # Overridden to work around https://github.com/django-guardian/django-guardian/issues/146,
+        # an issue with regards to guardian's PermissionMixin working with CreateView -
+        # where it requires an object even in a CreateView.
+        return None
```
cXhristian: Maybe we should create some generic mixin for this?

cXhristian:
Looks good. :shipit: 

Maybe we could use a better word than "chunks"? I'm not sure what that would be though.

apps/dashboard/chunks/dashboard/views.py
sklirg: We could voice in on the relevant issue and hope for a fix. :-) If we can fix it ourselves we can PR it as well.

sklirg:
We already use the package named chunks, so I kept the name for consistency. If we want to replace django chunks (#1376) we can rename it at that time.

PR #1539, opened 2016-04-22
Let django reverse urls to dashboard modules

sklirg:


templates/dashboard_base.html
```diff
@@ -83,7 +83,7 @@ <h4 class="center-text"><small>Velkommen</small><br />{{ request.user.first_name
                         <ul class="treeview-menu">
                             <li><a href="{% url 'user_list' %}"><i class="fa fa-angle-double-right orange"></i> Brukerprofiler</a></li>
                             {% if request.user.is_superuser %}
-                            <li><a href="/dashboard/auth/sso/"><i class="fa fa-angle-double-right orange"></i> Ekstern tilgang</a></li>
+                            <li><a href="{% url 'dashboard:index' %}"><i class="fa fa-angle-double-right orange"></i> Ekstern tilgang</a></li>
```
cXhristian: This will reverse to /dashboard/. I see that there is a namespace conflict for dashboard and dashboard sso. 
I think sso should have the namespace dashboard-sso.

myth: Agreed

sklirg: No,

```
>>> reverse('dashboard_index')
'/dashboard/'
>>> reverse('dashboard:index')
'/dashboard/auth/sso/'
```

sklirg: Though I agree with your second point. The reason I'm not changing it in this PR is that it's out of this PRs scope, and it may (or may not) break functionality tied to the sso module.

cXhristian:
Looks good.

PR #1558, opened 2016-04-28
Webshop nitpicking and fixes

### LINKED ISSUE #1338: Webshop nitpicking
- [ ] Overview over users orders.
- [ ] Save button in product add view should be under as well so the user does not have to scroll back upp to save.
- [x] Breadcrumbs does not work.
- [ ] Refund logic
- [x] Dashboard overview over payed products that Prokom needs to order.
- [ ] Send notifications if users have unpaid orderlines where products have a deadline.
- [x] If a size has stock display it.
- [x] If a selected size has a stock and is empty disable the "add" button
- [ ] Dashboard support for sizes
- [x] Reduce stock on purchase (Both total and ProductSize)
- [ ] Fix product list on mobile.
- [x] Add list of products in the confirmation email.
- [ ] Fix gallery functionality so images works properly for the webshop.
- [ ] Properly implement deadline handling.
- [x] Product description is duplicated on regular desktop view (e.g it is rendered twice).
- [ ] Mobile view content is generally not good. Most of the content overlaps and floats do not break correctly.
- [x] Top bar menu item Webshop overflows and is interpolated with dropdown user menu on medium width
- [x] There is an issue with reverse matching in dashboard (probably related to slug generation, ref stacktrace)
- [x] Breadcrumbs look terrible, use bootstraps pre-built breadcrumbs for navigation, and the h3 for titles as is tradition.
- [x] The `,-` means there is no remainder and it is a flat price. Either `150,-` or `150,00kr`.
- [ ] Add functionality to reduce amount in cart without deleting entire order line.
- [ ] `apps.payment.views:145 Name bought OrderLine object for 120.00 kr` is not a very good log entry. Needs moar ID's, hash, textual description, CSV of product ID's or something.

cXhristian:
#1338


apps/webshop/models.py
```diff
@@ -23,6 +23,22 @@ class Product(models.Model):
     deadline = models.DateTimeField(null=True, blank=True)
     active = models.BooleanField(default=True)
 
+    def calculate_stock(self):
+        if self.product_sizes.count() > 0:
+            return sum((product_size.stock for product_size in self.product_sizes.all()))
```
myth: Redundant parentheses? Generator expressions inside a method argument does not need wrapping.

apps/webshop/views.py
```diff
@@ -62,22 +91,31 @@ def post(self, request, *args, **kwargs):
...
+
         form = OrderForm(request.POST)
         if form.is_valid():
             order_line = self.current_order_line()
 
             size = form.cleaned_data['size']
+            quantity = form.cleaned_data['quantity']
+
+            if not product.enough_stock(quantity, size):
+                messages.error(request, "Det er ikke nok produkteter på lageret.")
```
myth: Misspelled "produkteter"

apps/webshop/models.py
cXhristian: Ah, didn't know.

myth:
Code looks good (havent tested it locally yet).
One thing I have to say tho, is I would really like to se proper docstrings on all the methods.

If using PyCharm or a plugin for your editor it's literally just pressing 3 " in a row and enter to generate the stub. Some day it would be awesome to be able to generate a sphinx documentation set from the code.

At least just being able to `ctrl+q` (In PyCharm) to get the method specification (and it's even more awesome with 3.5 and type hinting) and not having to browse the files yourself improves efficiency quite a bit. It is something I would like everyone to start doing.

apps/webshop/models.py
myth: It's really only necessary to wrap a generator expression in parenthesis if you are solely assigning it to a variable like `top = (kek.bur for kek in iterable)`

apps/payment/views.py
```diff
@@ -129,6 +129,11 @@ def webshop_pay(request):
             try:
                 stripe.api_key = settings.STRIPE_PRIVATE_KEYS["prokom"]
 
+                if not order_line.is_valid():
+                    error = "Ordren er ikke gyldig."
+                    messages.error(request, error)
+                    return HttpResponse(error, content_type="text/plain", status=500)
```
sklirg: HTTP 500 for an invalid order is a bit much, eh? The standard HTTP 400 BAD REQUEST should be fine. (An invalid order would be a client mistake, not a server side one, right?)

apps/webshop/models.py
```diff
@@ -23,6 +23,22 @@ class Product(models.Model):
...
+
+    def in_stock(self):
+        stock = self.calculate_stock()
+        return stock is None or stock > 0
+
+    def enough_stock(self, quantity, product_size=None):
+        stock = self.stock
+        if product_size:
+            stock = product_size.stock
+        return stock is None or stock >= quantity
```
sklirg: What is it that you want here? `return stock is None or ...` would return `True` if stock _is_ None, right? If we have `None` as stock, does that mean that we have unlimited? Or nothing?

apps/payment/views.py
cXhristian: I agree.

apps/webshop/models.py
cXhristian: If stock isn't specified then it is assumed that there is an unlimited supply yes.

sklirg: :+1:

Maybe add a line to model field help_text regarding this :-)

sklirg:
Looks good, commented on some of the stuff. 

I still want tests for webshop functionality, though, as it has to do with actual money. Yet again, this app is being rushed for release - so I don't expect it for now, but I think we should get this in place for payment related apps and views.

cXhristian:
PR updated. Tests will have to be written later. :crying_cat_face:

sklirg:
:+1:

PR #1589, opened 2016-06-23
Fix #1581. Add itex mail to mail participants

### LINKED ISSUE #1581: Add itex to "Send mail to participants" list

sklirg:


duvholt:
I'm not a fan of the arbitrary numbering we're using for emails, but I guess that's something that can be fixed in the upcoming group update. 
That's irrelevant to this pr though. Looks good.

PR #1606, opened 2016-07-24
Fix 1604: Add weekdays to Splash calendar

sklirg:
screenshot 2016-07-25 09 58 23

Would be nice if someone with a bit more CSS skills than me could verify that the `margin-top: -6px` isn't the worlds worst hack.

### DIFF
--- a/files/static/css/splash/main.css
+++ b/files/static/css/splash/main.css
@@ -285,7 +285,7 @@ article.cal {
   text-align: center;
   display: block;
   color: #ff7800;
-
+  margin-top: -6px;
 }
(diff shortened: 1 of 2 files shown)


duvholt:
In my opinion it would look better if you made the boxes bigger instead. 
Right now it looks a bit too compact. 

How about 72px height on li and a bit less margin on `.ca-left`?

OptimusCrime:
Give more padding please

sklirg:
screenshot 2016-07-25 09 58 23

So this should be good.

PR #1608, opened 2016-07-27
Fixed #1607 by using the flex method and moving inline-style to main.css

### LINKED ISSUE #1607: Layout bug when shrinking width or height to less than content on desktop
When the page reaches a width of < 960px (described as a static width in the main.css) or height < content height it hides some of the content. This is because the page uses this left: 50%; transform: translate(-50%, -50%) "hack", but since flex-box has come way longer the past years, this can be solved much cleaner now.

Here is the page now (notice the scrollbars):

image

niklasmh:
Moved one dirty inline-style liner from online-notifier.js into the main.css-file and made a class to refer to instead. Also replaced this hack:
`#container {
top: 50%;
left: 50%;
transform: translate(-50%, -50%);
}
`
With some flexbox like this:
`html {
	height: 100%;
}
body {
	min-height: 100%;
	display: flex;
	flex-direction: column;
}`
Which works very smooth when it comes to centering layout. I really recommend reading further into flex-box!

Scrollbars before:

image

The scrollbars now:

image

### DIFF
--- a/files/static/css/splash/main.css
+++ b/files/static/css/splash/main.css
@@ -29,13 +29,9 @@ a:hover {
   color: #2f6397;
 }
 #container {
-  width: 960px;
-  position: absolute;
-  top: 50%;
-  left: 50%;
-  -webkit-transform: translate(-50%, -50%);
-      -ms-transform: translate(-50%, -50%);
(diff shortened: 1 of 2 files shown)


sklirg:
Seems legit

PR #1616, opened 2016-08-07
Update min length limit in event fields

### LINKED ISSUE #1578: Shorten required length for some event fields
This is to accomodate for events planned long into the future, where all information may not be known at the time of event creation. However, we still want to have a lower limit - as the front page would look ugly if there was not enough text to ~fill the "featured" events' text boxes.

tmn:
I've set min length of short and normal ingress to 25, and description to 45. Because:

- A very valid ingress could be as short as:
  - Gjør deg klar til X-Fest!
  - Velkommen til kurs med ELV
  - Velkommen til bedriftspresentasjon med NIX

Often you don't have the description ready for an event. And a valid description could be:
- Mer informasjon kommer når [event] nærmer seg!

The form could enforce everyone to write something about the co-organizers etc. But the issue is more "the look on the front page" than the content of the page. How to present an event without much content (waiting for information etc.) could be discussed as a routine for the working committees, but I don't think the system should enforce it.

This PR fixes #1578

### DIFF
--- a/apps/events/admin.py
+++ b/apps/events/admin.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 from django.contrib import admin, messages
-from django.core import validators
 from django.utils.translation import ugettext as _
 from reversion.admin import VersionAdmin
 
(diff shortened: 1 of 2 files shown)


sklirg:
I agree. No matter how many times we ask for more information for the events, it simply is not always possible. Whenever this becomes a problem, the workaround is to add a lot of spaces until the max length is reached, which ends up looking ugly in the front end anyways.

Let's stop making this a nagging problem for our fellow committee members.

PR #1618, opened 2016-08-07
Upgrade to Bootstrap 3.3.7

### LINKED ISSUE #1447: Upgrade Bootstrap
Oops, looks like our Bootstrap version is almost two years old!
3.3.6 is the latest as of March 2016.

Bootstrap 4 is still in alpha and likely won't be released for a while so upgrading to a new 3.x release is worth it in my opinion.

tmn:
Most noticable changes would be the default blue color in bootstrap (why are we using their colors instead of own blue color?) - well, the new blue color is a little darker. And minor changes to the chevron. Please check any other stuff if you know uses weird bootstrap features.

Our code is depending on bootstrap variables which is a bad thing. We should create our own style guide instead of using their values. Further work on the styling would include decoupling from bootstrap values.

This PR is just an upgrade of the old bootstrap lib, and closes #1447


duvholt:
:+1: to decoupling from Bootstrap.

sklirg:
Let's create a separate issue for that. :+1:

PR #1629, opened 2016-08-14
Slack inviter

duvholt:
Slack invitation app with a simple API.
POST to `/api/v1/slack` with `email`.

This was created to enable simple signups from the splash, but we should probably add a form somewhere on the main site too.


myth:
:+1:

PR #1632, opened 2016-08-15
Add support for sending an email to application approvers.

sklirg:


apps/approval/appconfig.py
```diff
@@ -4,3 +4,7 @@
 class ApprovalConfig(AppConfig):
     name = 'apps.approval'
     verbose_name = 'Approval'
+
+    def ready(self):
+        super().ready()
+        import apps.approval.signals  # flake8: noqa
```
myth: Have you actually used type hinting comment syntax here? :D

apps/approval/signals.py
```diff
@@ -0,0 +1,27 @@
...
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import MembershipApproval
+from .tasks import send_approval_notification
+
+
+@receiver(post_save, sender=MembershipApproval)
+def new_membership_approval_handler(sender, instance, created, **kwargs):
+    """
```
myth: I see you use docstrings with type declaration, take it easy on my nerdboner.

myth:
Looks good

apps/approval/appconfig.py
sklirg: This is to prevent flake8 to report this as an unused import.

> lines that contain a # noqa comment at the end will not issue warnings.

Source: https://pypi.python.org/pypi/flake8

myth: Cool, then flake8 uses the comment syntax for type hinting. Guido and Co just released another PEP with actual type hinting syntax based on reserved syntax that will replace the comment syntax. Just thought it was interesting that they used a language feature that was not in standard use.

PR #1642, opened 2016-08-22
Do not crash if google plugin can't load Maps data

nixolas1:
Blocking google maps requests caused the init script to crash.
Added try/catch clauses around map loading code, because simply checking if the 'google' variable was undefined or not triggered the error as well.


sklirg:
looks good

Issue #1665, opened 2016-09-06
Do not fadeout alert error messages


duvholt:
At the moment all error messages to the user fade out after a few seconds.
This is okay for normal alert messages, but errors should stay on the screen. The user can hide them manually if they want.

sklirg:
I suggest we add functionality which will tell if the alert should be automagically dismissed.

- User logging in will trigger alert-success "User successfully logged in", which can be hidden automagically.
- User paying for an event will trigger alert-success "You have successfully paid for this event", which we should let stay on the screen

I'm not sure about the policy we've incorporated for warnings and errors, but we can probably let them stay on the screen without having to worry. 

Since we have alert-success' that we _want_ to auto-dismiss, and we have alert-success- that we _don't want_ to auto-dismiss, I suggest we add conditional in the dismiss-job that checks for a data property which will tell us if the alert should be automagically dismissed and do so, otherwise, don't dismiss it.

To automagically dismiss the alerts we _want_ to dismiss, we have to add the data-property to those events. For anything else, we don't have to do anything. This means we can add the functionality without having to do any refactoring - but rather add the auto-dismiss property on elements at any point in the future.

PR #1672, opened 2016-09-07
Fix alert fadeout

Aborysa:
#1665


Aborysa:
Contains errors in code. should be $(elm) instead of elm in init.js

Issue #1654, opened 2016-08-31
Filter events based on what the user can attend to


niklasmh:
Now you can see every single event, even those you can not attend to. It would be much easier to have an option to filter those out.

duvholt:
Fixed in #1673

PR #1673, opened 2016-09-07
Event search filter by access

Aborysa:
#1654

### DIFF
--- a/apps/events/views.py
+++ b/apps/events/views.py
@@ -191,6 +191,9 @@ def search_events(request):
 
     return render(request, 'events/search.html', {'events': events})
 
+def _can_attend(user,event):
+    return (not(event.is_attendance_event()) or 
+            event.attendance_event.is_eligible_for_signup(user))
 
 def _search_indexed(request, query, filters):
     results = []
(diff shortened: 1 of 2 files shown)


templates/events/index.html
```diff
@@ -84,6 +84,12 @@
             Mine arrangementer
           </label>
         </div>
+        <div class="checkbox col-xs-12 col-sm-12 col-md-12">
+          <label>
+            <input type="checkbox" name="optionCheckboxs" value="option1" id="attendable">
+            Arrangement du ikke kan delta på
```
duvholt: Doesn't it make more sense to invert this?

apps/events/views.py
```diff
@@ -213,7 +215,11 @@ def _search_indexed(request, query, filters):
     display_events = set()
 
     for event in events:
-        if event.can_display(request.user):
+        if (event.can_display(request.user) and
+            not(event.is_attendance_event() and
+                filters['attendable'] and
+                event.attendance_event.is_eligible_for_signup(request.user))):
+
```
duvholt: I think you should make a function that returns true or false to make this more readable.

Issue #1635, opened 2016-08-17
Save NTNU access card information to user profile


sklirg:
If we save the correct number(s) located on the student ID cards (the lower left one on the "old" cards or the one prefixed by EM on the "new" cards), we can generate the RFID (@duvholt). We can also reverse the RFID back into the code on the back of the cards.

This way, we can store the code(s) easily online and generate a printout for when we need to give NTNU Byggsikring these codes, rather than telling everyone to "write down this and that code and send it in an email to someone".

torjusti:
This post is mostly to recap this issue and what we want to do with it.

On the back of NTNU cards there is a code which is unique for each card. We often need to send these codes to NTNU Byggsikring to gain access to rooms and such, and therefore it would be nice if we could generate lists containing these using data already stored in onlineweb4. It would also be nice if users can use this to essentially set their own RFID code instead of registering directly at their first event, which I'd imagine slows the attendance keeping process a bit.

In my opinion, the RFID field and this field should be syncronized so that the fields are always consistent. To do that, we need to convert the codes to RFID when changing the code, and the RFID to the code representation when changing the RFID. I don't know if we want to expose both fields, or essentially have the RFID be some kind of computed field? I'd imagine we'd still want to get RFID data from regme, but perhaps only one should ever be exposed to the actual user as they're both essentially different representations of the same information. Thoughts?

There's a decent amount done on this issue in #1683 already, but I think we need a small discussion on how we want to finish it off?

sklirg:
I'd suggest storing the value on the back of the cards (find out which one), which can be computed to the RFID on demand using the information [here](https://github.com/dotkom/onlineweb4/pull/1683#issuecomment-247165268). This makes it easier for people to update the code themselves if required. It would also make it possible to compute the other NFC value (if there's an algorithm for that?) if we want to transition to the other chip in the cards, e.g. for a mobile (app/website) version of "regme", removing the need for a computer present at the events.

For event registration, post the RFID to the API, the API converts the RFID back into the code and does a lookup on that code. Same for Nibble. Or do the conversion client-side. Or support both.

oleast:
This information is stored on the user profile, and it is possible to change it via the code on the back of your card in settings in the new front-end.

PR #1683, opened 2016-09-14
Save NTNU access card information #1635

teatimes:
Lagt til felt og feilsjekking (7 eller 10 sifre)


duvholt:
Converting from number on the back of NTNU card to output from RFID scanning:

```python
# Convert from integer to binary padded to 32 bits
binary = bin(number)[2:].zfill(32)
# Split binary into bytes
bytes = [binary[i:i + 8] for i in range(0, len(binary), 8)]
# Reverse byte groups
bin_reversed = ""
for byte in bytes:
    bin_reversed += byte[::-1]
# Convert back to integer
rfid_binary = int(bin_reversed, 2)
# Pad to length 10
rfid = str(rfid_binary).zfill(10)
```

MrHanSan:
This PR will be closed because of conflicts and age.

PR #1695, opened 2016-09-23
Make the first line of the mail not appear twice on payment events

### LINKED ISSUE #1694: Mail for waitlist bumps on payment events gets duplicate first line
The first line in the mail for a waitlist bump on a payment event is added twice. 
Ex: 
> Du har stått på venteliste for arrangementet "Blåtur" og har nå fått plass.
> Du har stått på venteliste for arrangementet "Blåtur" og har nå fått plass.
> Dette arrangementet krever betaling og du må betale innen 2 dager.
> Prisen for dette arrangementet er 100kr.

nornes:
Fix #1694


nornes:
I wasn't able to test, but I can't see how this would break anything.

PR #1705, opened 2016-10-01
Fix #1704. Shrink statistics table and grow cake diagram

### LINKED ISSUE #1704: Feedback cake doesn't render if the names are very long
This happens if you open the page

pasted image at 2016_10_01 16_11

If you zoom out and reload, this happens
pasted image at 2016_10_01 16_20

sklirg:
It now looks like this instead, much more room for the glorious cake 🍰 .

pasted image at 2016_10_01 16_22


duvholt:
sick work m8
💯 💯 👌

PR #1708, opened 2016-10-05
Email when membership approved #1635

teatimes:


apps/approval/models.py
```diff
@@ -30,6 +32,22 @@ class Approval(models.Model):
     approved = models.BooleanField(_("godkjent"), default=False, editable=False)
     message = models.TextField(_("melding"))
 
+    def alert_user():
+        if approved:
+            header_message = "Ditt medlemskap i Online er godkjent"
+        else:
+            header_message = "Ditt medlemskap i Online er ikke godkjent"
```
sklirg: I think we can keep the header free of "spoilers", so that you have to open the email to read what it contains.

apps/approval/models.py
```diff
@@ -30,6 +32,22 @@ class Approval(models.Model):
...
+
+        email = applicant.get_email()
+        if (not email == None):
+            send_mail(
+                header_message,
+                header_message,
```
sklirg: The message could be different from the header :-)

It should also include the comment regarding why the membership was denied.

apps/approval/tests.py
```diff
@@ -63,3 +64,18 @@ def testIsFOSApplication(self):
         self.assertEqual(self.approval.is_fos_application(), False)
         self.approval.new_expiry_date = None
         self.assertEqual(self.approval.is_fos_application(), False)
+
+class EmailTest(TestCase):
```
sklirg: This test tests that Django's own mail function works. Django already tests that.

What you should do to test this behaviour is to test the view and send some data to it to verify that after someone clicks Accept/Deny, a mail is sent.

sklirg:
Wow, GitHub is harsh with "request changes" :-(

A huuuge cross.

Oh well, let's change it and this can probably be approved.

However, I'd like for the mail to be sent on an update of a model instance, rather than a specific view – this way the handling is consistent across Django Admin, Onlineweb4 Dashboard as well as eventual database migrations or future APIs (read about Django Signal handling and see the signals.py file which already does this for approval notification for administrators :-))

sklirg:
If you click "details" on the failing test, you can see that it's actually "isort" that fails. Isort is a tool we use to alphabetically sort imports, and has nothing to do with the actual tests. If you fix those issues (alas: sort the imports alphabetically and remove any imports not in use - hint: type `isort -rc apps/approval` to automatically sort the imports in apps/approval), you should see the output of the actual tests instead.

sklirg:
I merged this with develop and tested a bit.

When I applied for membership using the website -> Min side -> Medlemskap, I did not receive an email. So I looked a bit into this and quickly fixed it. 

I did two small fixes:

- Target "MembershipApproval" model when connecting signal (f227258)
- Better check for when to send email (only after `application.processed` is `True`, otherwise it would send an email saying application denied when you apply, which is silly) (56173fa)


The tests looks good. Nice work!

However, it seems like your last changes did something weird to the imports, so please sort those out. :)

Finally, I'd love to see the possibility to disable this easily (or enable..) using django settings. Check out the other approval task for how this is done.

PR #1716, opened 2016-10-07
Dashboard events

sklirg:
Create an event part 1
screenshot 2016-10-07 19 05 55

Create an event part 2
screenshot 2016-10-07 19 06 11

Detail / update view
screenshot 2016-10-07 19 10 41

Add company
screenshot 2016-10-07 19 10 50

Add attendance event
screenshot 2016-10-07 19 11 35

Attendance event detail view
screenshot 2016-10-07 19 11 54

Attendance event detail / update view
screenshot 2016-10-07 19 12 21

### ToDo

- [x] Change image source for events on frontpage
- [x] Add payment relation view

### Final notes

Barebones Django with templating does not support the kind of fancy ajax updating and data updates that we want to do in a simple way. That means this solution is not in any way the optimal, and it's not very nice. However, it could have been a lot worse – and it's currently **long** overdue. 

Using jQuery to do this will become a giant mess and be very ugly. I suggest we start ASAP on implementing this as yet another React-module in which we can swap out feature by feature with React components.

---

## Showstoppers

- [x] Bugs listed above
- [x] Code review
- [x] Local testing


duvholt:
:tada:

myth:
:clap: :champagne:

Will have a look when I get the time..

apps/dashboard/widgets.py
```diff
@@ -107,10 +107,16 @@ def render(self, name, value, attrs=None):
         else:
             final_attrs['value'] = ''
 
+        if attrs.get('placeholder', False):
```
nornes: I think the placeholder bug lies here. You aren't testing if a date is already set, so the placeholder is put in if it exists.

sklirg: Sounds about right. I'll have a look at this one time. 💃

sklirg: This is not it. And even if it was, it shouldn't have mattered since if a field has a value, that will have higher priority than the placeholder of the field. The placeholder is hidden if the field value is set.

sklirg: This was resolved in 9c0bc64.

Torrib:
Things that should be fixed before this is deployed to production:

- [ ] "Marks has been set" field should not be displayed in attendance
- [ ] "Active" field should not be displayed in feedback
- [ ] Should require at least one price to exists to be able to add a payment. (Payments without prices might cause some bugs).
- [ ] Event restriction is missing (Julebord upcoming)
- [ ] Unable to edit attendance after it has been created?
- [ ] Removing feedback caused a 404
- [ ] Error when manually adding a user
- [ ] "Show event" (Vis knappen) caused an error.
- [ ] Removing a company caused a 404

I was not able to test most attendee related functionality.
I have created a separate issue with more general feedback and less important things here: #1727

PR #1719, opened 2016-10-12
Create new API for attendance events and attendees

sardred:
Adds an endpoint for Attendees.
Updates a seemingly unused AttendeeSerializer and UserSerializer
Added (an easier) way to check if an attendee is on a waitlist.
The main magic here happens in the AttendViewSet that takes post requests from regme and handles the end user accordingly.

Sorry for the poorly organized commits. :disappointed:


sardred:
This might not be done yet. We're not quite sure how we're going to get the events to begin with, so we might require some changes there if we can't find a clever way to filter our way to what we want.

sardred:
I added filtering by attendance event. Everything should be complete now, and (hopefully) ready to merge.

sardred:
Okay, THIS time it's ready.

apps/events/views.py
```diff
@@ -319,7 +322,58 @@ class AttendanceEventViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin,
...
+        waitlist_approved = request.data.get('approved')
+        attendee = None
+
+        if username is not None:
+            try:
+                user = User.objects.get(username=username)
+                user.rfid = rfid
+                user.save()
+            except User.DoesNotExist:
+                return Response({'message': 'Brukernavn finnes ikke. Husk at det er brukernavn på Onlineweb! '
```
sklirg: Could we use `online.ntnu.no` instead of `onlineweb`?

sardred:
For real this time though.

PR #1733, opened 2016-10-19
Added regme.readwrite to scope

Aborysa:


sklirg:
not sure if we need this, but okay

PR #1739, opened 2016-10-20
Don't auto dismiss all bootstrap messages

### LINKED ISSUE #1665: Do not fadeout alert error messages

sklirg:
Fixes #1665.

### DIFF
--- a/files/static/js/init.js
+++ b/files/static/js/init.js
@@ -67,11 +67,7 @@ var initGoogleMaps = function() {
 // Fadeout alerts if they have the data-dismiss property
 var timeOutAlerts = function() {
   setTimeout(function () {
-    $('.alert').each(function() {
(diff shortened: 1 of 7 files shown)


files/static/js/init.js
```diff
@@ -64,14 +64,22 @@ var initGoogleMaps = function() {
     return map;
 }
 
+// Fadeout alerts if they have the data-dismiss property
+var timeOutAlerts = function() {
+  setTimeout(function () {
+    $('.alert').each(function() {
+      if ($(this).attr('data-dismiss')) {
```
OptimusCrime: Why not use the data tag in the selector? No need ro loop the alerts first

myth:
:shipit:

PR #1740, opened 2016-10-20
Add warning if trying to attend events with registration.end in the past

sklirg:
Requires #1739.

Otherwise, the bootstrap alert is probably faded out before the modal is opened.


myth:
:shipit:

PR #1751, opened 2016-10-26
Makes name more accessible from attendee, some fixes

sardred:
Some fixes include removing a filtering field on User that was based on a field that was removed earlier, changes a http response status code to something that doesn't break, and fixes some strings that were arrays for a silly reason.


OptimusCrime:
Looks good to me!

PR #1753, opened 2016-10-26
Event signup display right away

OptimusCrime:
If the signup for an event has yet to open, the orange box currently displays when signup opens. Like this:

skarmavbild 2016-10-26 kl 21 56 37

I know that the signup is listed in the info box to the left, but as an experienced Onliner your eyes will usually wander towards the orange box upon load. Currently, if the event has yet to open, you will have to start looking for the correct date in the info box to figure out when the event signup opens. I suggest altering the current message to simply stating when the event opens. This removes the need to look for the signup in the info box.

Purposed change:

skarmavbild 2016-10-26 kl 22 13 44


OptimusCrime:
FUCK

PR #1755, opened 2016-10-26
Add date to info box

OptimusCrime:
If the signup for an event has yet to open, the orange box currently displays when signup opens. Like this:

skarmavbild 2016-10-26 kl 21 56 37

I know that the signup is listed in the info box to the left, but as an experienced Onliner your eyes will usually wander towards the orange box upon load. Currently, if the event has yet to open, you will have to start looking for the correct date in the info box to figure out when the event signup opens. I suggest altering the current message to simply stating when the event opens. This removes the need to look for the signup in the info box.

Purposed change:

skarmavbild 2016-10-26 kl 22 13 44

### DIFF
--- a/templates/events/details.html
+++ b/templates/events/details.html
@@ -200,7 +200,13 @@
                                                 <p class="status-text">{{ user_status.message }}</p>
                                             {% endif %}
                                         {% else %}
-                                            <p class="status-text">{{ user_status.message }}</p>
+                                            <p class="status-text">
+                                                {% if user_status.status_code == 501 %}
+                                                    Påmeldingen åpner {{ attendance_event.registration_start|date:"d. M Y H.i" }}
+                                                {% else %}
+                                                    {{ user_status.message }}
+                                                {% endif %}
+                                            </p>
                                             {% if user_status.status_code == 403 %}
                                                 <p><button type="button" class="btn btn-primary" data-toggle="modal" data-target="#user-status-modal">Mer info</button></p>
                                                 <div id="user-status-modal" class="modal fade" role="dialog">


sklirg:
I think duplicating information can become confusing ("why is the same date listed twice?").

Additionally, the "top text" is used for displaying the "state" of the event's registration. It could say any of the following; "Du er meldt på dette arrangementet", "Du må være medlem for å delta ...", "Du er ikke i et klassetrinn som har tilgang ...". 

The "lower text" however will always tell you when the end of registration is/was.

Could we change it so that the lower text contains the time of registration start up until the actual registration start, then show unattend deadline (like now)?
This actually removes the information about the deadline for unattending, which might be interesting for some. 
This could be solved by having information about the unattend deadline available as a datatext for the lower text, kind of like we do for [when users are not allowed to attend](https://github.com/dotKom/onlineweb4/blob/63ae77931b9e355b467f729f219bb2654f8b1ca9/templates/events/details.html#L205).

Whatcha think?

OptimusCrime:
I did not change any of the internal logic, so this text will not override anything except what is illustrated in the images. As you can see I only apply this if the status code is 501. This maps directly to this line only: https://github.com/dotKom/onlineweb4/blob/develop/apps/events/models.py#L646

The other alternatives are not affected. The text will only appear IF the registration is not postponed and is happening in the future. 

I was thinking about calculating when the registration will open, but there are too many cases with rules/grades/suspensions etc. for that to be easily implemented. As you might be aware of, the event + delay system is a mess. The code is partially in python in the models file, partially in the view function and partially in the template. Rewriting all that to include the correct time for when a registration opens is a huge task, and we might as well consider refactoring the entire system if we chose to do this.

I wanted to implement this change because I feel that the current solution is bad for user experience. "Påmeldingen har ikke åpnet". "Okey, fine, so when does it open?". "Avmelding er ...". "Wtf, I don't care about that now, when does the registration open ffs?". And then you have to look for the registration start. It does not help that there are three dates in the info blue box either. I also feel that adding the time for when the registration opens in "small text" is redundant, as "Påmeldingen er ikke åpnet" and "Påmeldingen åpner ..." is more or less the same thing, except one contains all the relevant information, and the other is redundant.

I suggest either:

- Accept this PR and display the date in the scenario that nothing is postponed.
- Rewrite the internal handling of orange box feedback to be more precise and user friendly (e.g. always include when the registration opens, even with delays included), including untangling like 2k lines of code.

There is no reason we can't do the first one now and plan the next one for later.

My 5 cents.

PR #1757, opened 2016-10-29
Bump django filebrowser to 3.7.2

sklirg:


sklirg:
3.7.2 apparently doesn't run well on py3.5 which these tests run with.

PR #1762, opened 2016-11-03
Add interface to see and list pending feedback schemas in profiles

OptimusCrime:
In an effort to make feedback schemas easier to deal with, I've implemented this relatively simple sub section in the profile.

Right now, the feedback schemas are completely decoupled from the actual site, meaning you can not access a pending (or answered for that sake) schema from within ow4. You'll have to use the link sent in the daily reminding emails.

It's implemented as a context processor. It checks if the current user has any active schemas pending and adds it in the user menu dropdown. 

skarmavbild 2016-11-03 kl 14 35 03

This will also reveal a new sub section in the profile where these schemas are nicely listed. 

skarmavbild 2016-11-03 kl 14 35 16

skarmavbild 2016-11-03 kl 14 35 23

### DIFF
--- a/onlineweb4/context_processors.py
+++ b/onlineweb4/context_processors.py
@@ -5,6 +5,7 @@ from django.utils import timezone
 
 from apps.feedback.models import FeedbackRelation
 
+
 def context_settings(request):
     context_extras = {}
     if hasattr(settings, 'GOOGLE_ANALYTICS_KEY'):
@@ -12,6 +13,7 @@ def context_settings(request):
     if hasattr(settings, 'HOT_RELOAD'):
         context_extras['HOT_RELOAD'] = settings.HOT_RELOAD
     return context_extras
+
 
 def feedback_notifier(request):
     context_extras = {}
--- a/onlineweb4/context_processors.py
+++ b/onlineweb4/context_processors.py
@@ -1,4 +1,7 @@
+# -*- coding: utf-8 -*-
+
 from django.conf import settings
+from django.utils import timezone
 
 from apps.feedback.models import FeedbackRelation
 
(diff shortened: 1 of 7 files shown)


onlineweb4/context_processors.py
```diff
@@ -1,8 +1,42 @@
...
+    if not request.user.is_authenticated():
+        return context_extras
+
+    active_feedbacks = FeedbackRelation.objects.filter(active=True)
+    for active_feedback in active_feedbacks:
+        if active_feedback.content_object is None:
+            continue
+
+        # Making sure we have an end data, and that the event is over
+        # and that the feedback deadline is not passed (logic reused from apps.feedback.mommy)
```
sklirg: Is this copy pasta from feedback mommy? If so, could you make it a more generic function and call that one instead? So we don't have the same code two places and forget to update one of them in the future..

OptimusCrime:
Should perhaps include the text from the email that deadline is at midnight on the deadline day.

sklirg:
Just a quick  thought since this could be the start of a more hefty notification system.

Should we do this asynchronously instead, so that it happens after page load rather than during? If there are a lot of notifications to consider (approvals, feedback schemas, payments due, poster orders, rutinator tasks, ...) this could probably increase page render time by some time.

OptimusCrime:
A notification system would need a much more defined plan/structure than what I am able to do with this for now. The changes in this PR is pretty simple, and can easily be adjusted with the introduction of a fully fleshed out notification system afterwards. This PR was simply meant to introduce feedback functionality for now, while I knew it would need a rewrite once such a system is in place.

It can either be merged in now and we can start work on a notification system, or we can build the notification system and integrate this afterwards. The former could ship this feature now, while the latter would be some time next semester/whenever this system is ready.

The query used to filter out this is super light, and I'd argue that you can do many queries before it begins to impact the user noticeable. I am pretty sure integrating such a notification system could be done on load instead of async without a significant impact.

My five cents.

sklirg:
Since this is a feature we currently do not have anywhere else on the page I think it's fine to do it this way for now. We can jump back into analysis mode when we're about to implement a full notification system :)

Get the marks green again and I'll approve.

OptimusCrime:
Good! I'll fix.

PR #1765, opened 2016-11-08
Fix #1764

### LINKED ISSUE #1764: Check if user is attending before trying to attend user
If a user has a cached version of the website available with the "Attend"-button visible, and they attend from another device, then try to attend from the first device, you'll get a stacktrace due to unique constraint failed from the database, since the user already attends the event.

This should not be the case, and we should add a check for if the user is already attending.

sklirg:
This adds a failing test for the issue. Solve the issue and the test should become green!

### DIFF
--- a/apps/events/tests.py
+++ b/apps/events/tests.py
@@ -2,6 +2,7 @@
 
 import datetime
 import logging
+import os
 
 from django.conf import settings
 from django.contrib.auth.models import Group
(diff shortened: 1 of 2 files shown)


sklirg:
WTB drone :cry:

myth:
50k gold, and an epic mount plz

PR #1768, opened 2016-11-09
Returns the attended attendee on registration

sardred:
Unsure about the removal on line 348 here.


sardred:
Merging as approved by @sklirg

PR #1769, opened 2016-11-09
Frontpage events react

duvholt:
Do not merge yet!


duvholt:
Currently running on https://dev.online.ntnu.no

This is what it currently looks like:

PR #1777, opened 2016-11-16
Make dashboard sidebar more mobile friendly by positioning it fixed

OptimusCrime:
The dashboard was never intended for mobile users. However, because we use Bootstrap we are at the very least given some responsiveness for free. The dashboard sidebar has a problem where, if the actual page content is shorter than the sidebar, it will render incorrectly.

This PR tries to fix that. Instead of having the sidebar positioned absolute, it is fixed and overflowed automatically on the y axis. This creates a secondary scrollbar, but I see no other options to solve this.

See illustrations below.

Current behavior:
current

Suggested behavior:
new

Sidenote: The content was altered by the inspector tool to create a view that has a taller sidebar than content, however this may happen in real life scenarios as well. With the fixed approach, scrolling feels more natural on mobile units. Desktops are left more or less unaltered.

### DIFF
--- a/files/static/less/dashboard/main.less
+++ b/files/static/less/dashboard/main.less
@@ -213,11 +213,12 @@ body {
 }
 /*left side - contains sidebar*/
 .left-side {
-  position: absolute;
+  position: fixed;
   width: 220px;
   top: 0;
   bottom: 0;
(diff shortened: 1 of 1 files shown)


duvholt:
Looks good :+1:

PR #1779, opened 2016-11-23
Allows regme to attend attendees by username

sardred:
Also added some comments to try to make the big fat sack of crud that is AttendViewSet a bit less repulsive.

### DIFF
--- a/apps/events/views.py
+++ b/apps/events/views.py
@@ -346,6 +346,7 @@ class AttendViewSet(views.APIView):
         username = request.data.get('username')
         waitlist_approved = request.data.get('approved')
 
+        # If attendee has typed in username to bind a new card to their user
         if username is not None and rfid is not None:
             try:
                 user = User.objects.get(username=username)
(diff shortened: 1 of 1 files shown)


apps/events/views.py
```diff
@@ -356,23 +357,44 @@ def post(self, request, format=None):
...
+                return Response({'message': 'Brukernavnet finnes ikke. Husk at det er et online.ntnu.no brukernavn! '
+                                            '(Prøv igjen, eller scan nytt kort for å avbryte.)', 'attend_status': 50},
+                                status=status.HTTP_400_BAD_REQUEST)
+
+            # If attendee tried to attend by card, but card isn't tied to a user
+            else:
+                return Response({'message': 'Kortet finnes ikke i databasen. '
+                                            'Skriv inn et online.ntnu.no brukernavn for å '
+                                            'knytte kortet til brukeren og registrere oppmøte.',
+                                 'attend_status': 40}, status=status.HTTP_400_BAD_REQUEST)
```
sklirg: Will the user actually be attended after this (which is implied by the wording in `message`)? I can't see any code that tries to attend the user at this point.

Aborysa: `         try:
 -            attendee = Attendee.objects.get(event=event, user__rfid=rfid)
 +            # If attendee is trying to attend by username
 +            if rfid is None:
 +                attendee = Attendee.objects.get(event=event, user__username=username)
 +            else:
 +                attendee = Attendee.objects.get(event=event, user__rfid=rfid)`

`             attendee.attended = True
              attendee.save()`

Aborysa: @line-259

sardred: This would send them back to regme, where we another post request, this time with a username _and_ an RFID. This will send them through [this little tidbit](https://github.com/dotKom/onlineweb4/blob/f0d235d8c69db4998eec92ef246e126e4b5f7081/apps/events/views.py#L349-L358) and then attend as usual.

sklirg: Cool cool.

PR #1787, opened 2017-01-03
Major frontend code refactor

### LINKED ISSUE #1732: Make sure JS/CSS is ran through compressor
Compressor minifies files and make sure that changed resources like JS/CSS are not cached by the browser by using a hashed filename.

Files that contain `.min.js` are already minified and are likely libraries. No need to run them through compressor.

Example of compressor:
https://github.com/dotKom/onlineweb4/blob/c43de3acc27b6429352d88668b14d81ca85d6d2f/templates/gallery/index.html#L75

Example of file that needs to be fixed:
https://github.com/dotKom/onlineweb4/blob/d725049d5554b90adba718a1bafc98cecdc69646/templates/dashboard_base.html#L336

This has likely not been done correctly in many templates. 

Simple search for JS and CSS blocks:
https://github.com/dotKom/onlineweb4/search?utf8=%E2%9C%93&q=%22block+js%22++OR+%22block+styles%22&type=Code

duvholt:
This is the result of procrastination while studying for exams and a christmas vacation with a decent amount of free time. I originally planned to do this in smaller steps, but since our code was a bit of a mess with global functions and dependencies eveywhere it was easier to just convert everything.

The goal of this refactor was to make our frontend code easier to work with. Now we can easily split up big files and reuse util functions. Now we can also write EcmaScript 6 syntax everywhere. 

**There is a high chance that I have broken something in this pull request.** This needs major testing before we can use it in production. Every page that uses CSS and JS should be tested.

## Summary of changes

- Improve and document webpack config
- Build almost all static files (js, css, images etc) using webpack
- All code in the assets folder passes linting¨
- Remove django-compressor
- Create webpack entries for all our main apps
- Fetch frontend libraries from npm instead of local outdated copies
- Update frontend libraries
- Remove several unused libraries
- Rewrite some of the js components
- Split up and move Utils.js and Dashboard.tools functions to a common utils folder

### Exceptions

- jqPlot isn't on npm so I have added a copy in the assets/common and some webpack hacks to make it work with jQuery from npm
- I haven't done any major modifications to Stripe related code
  - Partly because I don't want to mess up any payments before we've tested things properly and because the system might warrant a rewrite (use standard Stripe button instead?)
  - jQuery and jQuery.cookie is manually included everywhere stripe is used for now
- Google Charts and Google Maps is still included using cdn 
- Added an ugly hack to load static files from python packages. Works pretty decent actually
- Some images left in `files/static`. Didn't bother to check if they were used.

Resolves #1732.


myth:
My lord, what have you done... 🏆

OptimusCrime:
Status: Review Needed. Files change: 467.

Well, alright then.

myth:
> 28 days later...

PR #1789, opened 2017-01-11
Add an initial Makefile

sklirg:
- `make` will pull, build and run OW4 using docker and docker-compose
- `make up` will run OW4 using `docker-compose up -d`, not re-building
anything.
- Check the readme or makefile for more information.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build up run logs start stop down rm status
+.PHONY: build up run logs start stop down rm status exec django manage
 
 build:
 	docker-compose pull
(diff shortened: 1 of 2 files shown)


sardred:
For whatever reason, the exec and manage commands of the makefile don't work. Running `make manage migrate` results in `make: *** No rule to make target 'migrate'.  Stop.`, suggesting migrate is the problem. However, running `make exec django python manage.py migrate` results in `Makefile:31: recipe for target 'exec' failed`. Running `docker-compose exec django python manage.py migrate` _does_ work, but having to run the whole command for it to work defeats the purpose of the Makefile. 
As an aside, should there be more clarity in the readme that one needs to run migrate (and probably createsuperuser) to really get started, or do we just assume that whomever pulls the repo knows some level of Django?

sklirg:
Will have a look at this.

PR #1790, opened 2017-01-13
Allow use of environment variables to override Django settings

sklirg:
This allows us to easily override Django settings using environment variables rather than `onlineweb4/settings/local.py`. Both works alongside each other, because if you override something in `local.py`, you'll override the configuration that gets a value from the environment. This means that you should not have to touch anything on your own if you want to continue using your current development configuration setup.

To override a setting, pass an environment variable to docker or docker compose, or add them to the `.env` file which is supplied to the Django container in Docker Compose.

* This makes use of [Python Decouple](https://github.com/henriquebastos/python-decouple) to enhance environment variables to configure Django settings (it makes it really easy to parse csv or booleans, which are strings when Python imports them).

* It also includes [DJ-Database-URL](https://github.com/kennethreitz/dj-database-url) which is a much simpler way to write database connection configuration. Simply pass `postgres://localhost:5432` as `DATABASE_URL` (or parse it in local settings) and it will automatically populate the appropriate fields in the Django `DATABASES` settings. This is for easy configuration using environment variables, especially for containerized deployments, so we don't have to ship in 5 different database parameters, we can do with one.


myth:
Good job, this is something that has been long overdue.

One thing though, are you sure we want to have the `.env` file in the project directory (and hence also in gitignore)?

Would it not be better suited in the parent dir's env directory as we currently have it, or in `/etc/default/` or somewhere else that does not allow someone having an accident and committing credentials to history?

Remember https://github.com/Samfundet/Samfundet/pull/8 ^^

sklirg:
Sure. I'll have a look at it. It might be better to pass a path to an ENV-FILE instead of expecting it to be in project root anyways.

sklirg:
I had a look at it, and the supported way of doing this is either using a shared env file (but that requires everyone to create it, which is tedious, and would be the same problem I'm trying to get rid of), committing this .env file to GH but proxying vars from host (see 0cd7975), or supply them in docker-compose in the same way.

I think the latter is the cleanest. Comments?

PR #1791, opened 2017-01-14
Enhancements for authentication and approvals dashboards

sklirg:


templates/auth/dashboard/user_detail.html
```diff
@@ -97,11 +101,11 @@
             <div class="panel-body">
                 <div class="col-md-6">
                     <h4>Allergier</h4>
-                    <p>{{ user.allergies }}</p>
```
duvholt: Use `{{ user.allergies|default:"-" }}` instead?
Might just be odd since most of the other fields need if blocks.

sklirg: Good idea.

PR #1796, opened 2017-01-25
Change password button

martinlunde:


OptimusCrime:
Provide print screens for stylish changes like this. Before and after is even better than just the result. Code looks good though.

PR #1797, opened 2017-01-25
Replace filebrowser

sklirg:
Sounded like a chore, actually worked out pretty nicely out of the box.


sklirg:
Because migrations still contain filebrowser, we'll have to squash them to remove the dependency on it (from requirements.txt). We should await doing this until we know we're safe to do so; by testing thoroughly. Nothing should have been using filebrowser for the past couple of months though, but you never know.

PR #1798, opened 2017-01-25
Only run webpack_resolve.py when debugging

duvholt:
Running it in production causes permission issues and is not needed.


sklirg:
We should probably add some docs somewhere some time that this command has to be run before a new production/staging environment is set up (docker containers...).

PR #1804, opened 2017-01-29
Fix Bootstrap overlay covering wiki modals

duvholt:


sklirg:
I'd love a screenshot, but I think I trust you enough to have fixed this.

PR #1805, opened 2017-01-30
[requires.io] dependency update on develop branch

dotkom-build:


requirements.txt
```diff
@@ -1,42 +1,42 @@
...
 pdfdocument==3.1
-Unidecode==0.4.19               # Translates every unicode symbol to the closest ascii. online_mail_usernames
+Unidecode==0.04.20               # Translates every unicode symbol to the closest ascii. online_mail_usernames
 django-markdown-deux==1.0.5
 django-formset-js==0.5.0        # used to dynamically create new formsets
-django-cors-headers==1.1.0      # Enable CORS for Nibble
+django-cors-headers==2.0.1      # Enable CORS for Nibble
 django-datetime-widget==0.9.3   # Datetime widget for forms
-django-webpack-loader==0.3.3    # Integration with webpack
+django-webpack-loader==0.4.1    # Integration with webpack
```
sklirg: @duvholt: Could you check if this is safe to upgrade?

https://github.com/owais/django-webpack-loader/compare/0.3.3...0.4.0

sklirg: bump @duvholt

duvholt: Shouldn't be a problem at all. We only use the basic functionality of django-webpack-loader.

PR #1807, opened 2017-02-01
Initial G Suite Membership Syncer

sklirg:


sklirg:
I've cleaned up a bit and connected it to User groups signals now. I've also tested it quite a bit on staging, and it looks like it works like it should :)

onlineweb4/settings/base.py
```diff
@@ -1,28 +1,39 @@
...
 from django.contrib.messages import constants as messages
+from webpack_resolve import create_resolve_file
 
 # Directory that contains this file.
 PROJECT_SETTINGS_DIRECTORY = os.path.dirname(globals()['__file__'])
 # Root directory. Contains manage.py
 PROJECT_ROOT_DIRECTORY = os.path.join(PROJECT_SETTINGS_DIRECTORY, '..', '..')
 
-TEST_RUNNER = "django_nose.NoseTestSuiteRunner"
+sys.dont_write_bytecode = config("OW4_PYTHON_DONT_WRITE_BYTECODE", cast=bool, default=True)
```
myth: Why do we need this? Has this caused us any issues lately?

apps/authentication/signals.py
```diff
@@ -2,22 +2,28 @@
...
-# from apps.authentication.models import OnlineUser
 from apps.authentication.tasks import SynchronizeGroups
+from apps.gsuite.mail_syncer.main import update_g_suite_user
 
 User = get_user_model()
 logger = logging.getLogger('syncer.%s' % __name__)
 sync_uuid = uuid.uuid1()
 
 
+def run_group_syncer(user):
```
myth: Use pycharm to mass-generate some pydoc for these functions perhaps?

apps/gsuite/mail_syncer/main.py
```diff
@@ -0,0 +1,51 @@
...
+logger = logging.getLogger(__name__)
+
+
+def insert_ow4_users_into_g_suite(domain, group_name, missing_users):
+    for missing_user in missing_users:
+        insert_ow4_user_into_g_suite_group(domain, group_name, missing_user)
+
+
+def remove_excess_g_suite_users(domain, group_name, g_suite_excess_users):
+    logger.info("Cleaning G Suite group '%s' (Removing %s)" % (group_name, g_suite_excess_users))
```
myth: Should we start using new style string formatting, or didn't that catch on?

apps/gsuite/mail_syncer/utils.py
```diff
@@ -0,0 +1,265 @@
...
+    user_key = get_user_key(domain, user)
+    logger.debug("Getting G Suite user membership list for '%s'." % user_key)
+
+    directory = setup_g_suite_client()
+
+    try:
+        groups = directory.groups().list(userKey=user_key).execute().get('groups')
+    except HttpError as e:
+        logger.error('HttpError when requesting user group membership list: %s' % e)
+        # @ToDo: Find out how to handle this properly.
```
myth: Have you found out how to do this properly yet?

myth:
This was a rather large PR, but LGTM from what i've seen.

Some general thoughts/pointers:

* Needs moar pydoc plz. Makes using the functions so much easier when in an IDE environment.
* Something something about new vs old style string formatting, but that's just nitpicking.
* GREAT JOB MAN, DOING GODS WORK HERE

onlineweb4/settings/base.py
sklirg: Copied it from example-local.py. No idea. Is it a Python 2.x thing? We could drop it I guess.

apps/gsuite/mail_syncer/utils.py
sklirg: Not really. Since it might be triggered on a user save action it should probably not raise an exception and rather continue on, so the user is actually saved. I'll look into it.

onlineweb4/settings/base.py
myth: You really only need to disable it if you don't have write permissions to the directory where the interpereter would write it, or use another featue of the bytecode config to write it somewhere else if it is being run from a network disk for instance.

The interpereter will generate new files if the `.py` file is newer than the bytecode file anyway.
It will help time to first byte after restarting the process, since every import does not have to be re-compiled. Not a big deal, but it is noticable when you restart the ow4 process.

apps/gsuite/mail_syncer/main.py
sklirg: Will fix.

apps/authentication/signals.py
sklirg: On it.

PR #1817, opened 2017-03-01
Frontpage react business

nornes:
Reactify the business section of the frontpage


sardred:
Closing because we left this open for a year, nobody knows the status, and if we actually want to write this low-priority over to React then we'll do it as a redesign without bootstrap.

PR #1830, opened 2017-03-21
Enable drone caching of npm and pip

sklirg:

- [x] QA

## Description of changes

Adding a plugin to drone to cache node_modules and pip to speed up PR builds.


sklirg:
From a random build, we saved 

* ~120 seconds on pip install (from 350 seconds to 120 seconds)
* ~80 seconds on npm install (from 90 seconds to 10 seconds)

This way only stores the pip cache, not the installed pip packages. Npm however, stores the installed packages.

The plugin requires "trusted" status on the repo, though, which might be a security concern. Would love some input.

PR #1842, opened 2017-03-30
Show SSO errors

duvholt:
## Code Checklist

## Description of changes

Show SSO errors instead of blank page.


duvholt:
The footer is still present on the page so isn't that covered by this?
image

I've only ever seen one error, but I think the only available information is the description.

And from what I can gather there isn't any information about the application that can be shown. 
Not without doing some manual checks and I have no idea how to do that.

PR #1845, opened 2017-04-19
deleted node_modules from docker-compose.yml, because of docker volum…

Westermoen:
…e conflicts

## Code Checklist

## Description of changes
read commit message


sklirg:
This is actually a sneaky hack we use to cache node_modules in docker without installing them in the local directory. It also reduces the overhead of sending the full node_modules folder to docker every time it builds.

PR #1851, opened 2017-04-26
Frontpage event filter logic change

niklasmh:
## Changing the logic of frontpage event types

- [ ] QA / Code Review

## Code Checklist

## Description of changes

Added more logic to which buttons is toggled on event types. This is because people mostly want to filter out all other options when they select one. When selecting all manually, the logic will notice and not deselect all again, but just toggle them as usual. People will mostly only click once and leave, so it is important to do the first step most intuitive as possible.

## Use cases in screenshots

First look:
image

Then, after selecting"Kurs":
image

After selecting all:
image

Then, after selecting "Kurs" again (after all was selected manually):
image

### DIFF
--- a/assets/frontpage/events/utils/index.js
+++ b/assets/frontpage/events/utils/index.js
@@ -1,58 +1,51 @@
 export const toggleEventTypeDisplay = (state, eventType) => {
-  let newState = state;
+  const newState = Object.assign({}, state);
+  const selected = [];
   let eventCount = 0;
-  let selected = [];
-  for (let key in state.eventTypes) {
+
+  for (const type of Object.values(newState.eventTypes)) {
     eventCount++;
-    if (state.eventTypes[key].display) {
-      selected.push(key);
+    if (type.display) {
+      selected.push(type.id);
     }
   }
 
   // If only one event is selected and is being deselected, then show all and set dirty to false
-  if (selected.length === 1 && eventType === newState.eventTypes[selected[0]]) {
-    newState.dirty = false;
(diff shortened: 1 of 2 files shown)


assets/frontpage/events/utils/index.js
```diff
@@ -1,10 +1,66 @@
-export const toggleEventTypeDisplay = (state, eventType) => (
-  Object.assign({}, state.eventTypes, {
-    [eventType.id]: Object.assign({}, eventType, {
-      display: !eventType.display,
-    }),
-  })
-);
+export const toggleEventTypeDisplay = (state, eventType) => {
+  let newState = state;
```
duvholt: If you want to clone an object use `Object.assign({}, state)`.
Right now the original state is also being modified.

assets/frontpage/events/utils/index.js
```diff
@@ -1,10 +1,66 @@
...
+      }),
+    });
+  }
+
+  // If none of the edge cases above
+  else {
+    newState.eventTypes = Object.assign({}, state.eventTypes, {
+      [eventType.id]: Object.assign({}, eventType, {
+        display: !eventType.display,
+      }),
```
duvholt: This code seems to repeat a lot. Maybe make a helper function that takes in eventType and display value?

duvholt:
And fix the eslint errors. 
Most editors have an eslint plugin that you can install.

niklasmh:
@duvholt Can you approve changes?

PR #1853, opened 2017-05-03
Feat contact

AslakHol:
Feature PR

## Code Checklist

## Description of changes
Added a contact us page. It can be found under the user menu

Old:
image

New:
image


apps/contact/views.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def contact_submit(request):
+    if request.method == 'POST':
+        form = ContactForm(request.POST)
+        if form.is_valid():
+            if form.cleaned_data['contact_receiver'] == '1':
+                choice = ['hovedstyret@online.ntnu.no']
+            else:
+                choice = ['dotkom@online.ntnu.no']
```
sklirg: This could be imported based on the list defined in `forms.py`, so that a change to one does not require a change to the other.

Likewise, by changing the structure of the list in `forms.py`, it's possible to do this matching automatically if more emails are to be added in the future.

apps/contact/views.py
```diff
@@ -0,0 +1,33 @@
...
+
+            subject = '[Kontakt]' + form.cleaned_data['contact_name'] + 'har kontaktet dere gjennom ow4'
```
sklirg: Use Python string substitution for this, looks much cleaner (also, by using concatenation by `+` here, you don't have spaces between the strings, so it'll be `"[Kontakt]Nameharkontaktet dere gjennom ow4"`.

Like this: `[Kontakt] {name} …`.

apps/contact/views.py
```diff
@@ -0,0 +1,33 @@
...
+            content = form.cleaned_data['content']
+            from_email = form.cleaned_data['contact_email']
+            to_email = choice
+
+            EmailMessage(subject, content, from_email, to_email).send()
+            messages.success(request, 'Mailen ble sendt')
+        else:
+            messages.error(request, 'Mail ble ikke sendt')
```
sklirg: Add steps to help the user to actually send the email.
Most likely this would be suggesting to try again or contact dotkom for help.

templates/contact/index.html
```diff
@@ -0,0 +1,68 @@
...
+                    	{{ form.contact_receiver.label }}
+                    </div>
+                    {{ form.contact_receiver|addclass:"form-control" }}
+                    <hr>
+					{{ form.contact_name|addclass:"form-control" }}
+					{{ form.contact_email|addclass:"form-control" }}
+					{{ form.content|addclass:"form-control" }}
+					<div id="captcha">
+						{{ form.captcha }}
+					</div>
```
sklirg: You can actually loop over these fields, so you don't have to repeat a lot of the code: https://docs.djangoproject.com/en/1.11/topics/forms/#looping-over-the-form-s-fields

templates/contact/index.html
```diff
@@ -0,0 +1,68 @@
...
+			<div class="col-md-6">
+				<form method="POST" id="contact_form" action="{% url 'submit' %}">
+					{% csrf_token %}
+                    <div id="choice_label">
+                    	{{ form.contact_receiver.label }}
+                    </div>
+                    {{ form.contact_receiver|addclass:"form-control" }}
+                    <hr>
+					{{ form.contact_name|addclass:"form-control" }}
+					{{ form.contact_email|addclass:"form-control" }}
```
sklirg: If the user is logged in, we can get these fields automatically.

sklirg:
Also, I'm not sure if we should require the email and/or name field. Some might not want to leave their contact details, and it's pretty easy to fake the fields if they don't want to.

apps/contact/views.py
```diff
@@ -0,0 +1,33 @@
...
+            else:
+                choice = ['dotkom@online.ntnu.no']
+
+            subject = '[Kontakt]' + form.cleaned_data['contact_name'] + 'har kontaktet dere gjennom ow4'
+            content = form.cleaned_data['content']
+            from_email = form.cleaned_data['contact_email']
+            to_email = choice
+
+            EmailMessage(subject, content, from_email, to_email).send()
+            messages.success(request, 'Mailen ble sendt')
```
sklirg: Add some logging here. Try to log user (username) if the user is logged in.

templates/contact/index.html
AslakHol: If we make the fields not required, which after thinking for a bit I think is a good idea, we don't want to do this. Keeping the fields there and changing the placeholder text to reflect the fields voluntary nature would make it clear to the user that you can leave an anonymous message even while logged in.

apps/contact/views.py
AslakHol: I don't think we want to do this if we want people to be able to leave anonymous messages.

AslakHol:
Biip boop, I think I've addressed your changes Sklirg

apps/contact/views.py
sklirg: I think we want to do this in case some people abuse it. The logs are only visible for dotkom.

apps/contact/views.py
```diff
@@ -0,0 +1,28 @@
...
+def index(request):
+    context = {'form': ContactForm}
+    return render_to_response('contact/index.html', context, context_instance=RequestContext(request))
+
+
+def contact_submit(request):
+    if request.method == 'POST':
+        form = ContactForm(request.POST)
+        if form.is_valid():
+            subject = '[Kontakt]' + form.cleaned_data['contact_name'] + 'har kontaktet dere gjennom ow4'
```
sklirg:
```
>>> data['contact_name'] = "Sklirg"
>>> subject = '[Kontakt]' + data['contact_name'] + 'har kontaktet dere gjennom ow4'
>>> print(subject)
[Kontakt]Sklirghar kontaktet dere gjennom ow4
```

I still think you want spacing here.

As I suggested previously, use string substitution, like this:
```
>>> subject = '[Kontakt] {name} har kontaktet dere gjennom ow4'.format(name=data['contact_name'])
>>> print(subject)
[Kontakt] Sklirg har kontaktet dere gjennom ow4
```

However -- since name is not required, it might be more useful for the subject to be "Ny kontaktførespørsel fra online.ntnu.no" (note: We try to discourage the use of "ow4", since not everyone know what "ow4" is. We prefer "onlineweb" or "online.ntnu.no").

AslakHol:
@sklirg I think I've addressed your requested changes.

templates/base.html
```diff
@@ -169,7 +170,7 @@
                     </div>
                     <div class="row">
                         <div class="col-md-12">
-                            <p>Har du funnet en feil på nettsiden?<br>Ta kontakt med <a href="mailto:dotkom@online.ntnu.no">Utviklingsteamet</a></p>
+                            <p>Har du funnet en feil på nettsiden?<br>Ta kontakt med <a href="{% url 'contact_index' %}">Utviklingsteamet</a></p>
```
sklirg: I'm all for emails, so we could leave this as a mailto link I think.

AslakHol:
I tested the logging on dev, so it should be good.

PR #1871, opened 2017-07-30
Make OW4 an OpenID Connect Provider

sklirg:

- [x] QA / Code Review

## Code Checklist

    - [x] I cannot apply the required changes in production before this is deployed.

## Description of changes

Makes OW4 an OpenID Connect provider so that we can create clients using OIDC rather than pure OAuth2.

The change required in production is to run `python manage.py creatersatoken` to create an RSA token for the OIDC server to provide.

### DIFF
--- a/templates/oidc_provider/authorize.html
+++ b/templates/oidc_provider/authorize.html
@@ -1,0 +1,42 @@
+{% extends "sso/index.html" %}
+{% block content %}
+<div id="sso-panel">
+{% if not error %}
+    <div id="sso-panel-header">
+        <h3>Ekstern autentisering</h3>
+    </div>
+    <div id="sso-panel-body">
+        <h4>Tjenesten {{ client.name }} ønsker tilgang til denne informasjonen om deg.</h4>
+        <form id="sso-authorization-form" method="post" action="{% url 'oidc_provider:authorize' %}">
+            {% csrf_token %}
+
+            {{ hidden_inputs }}
+
+            <p>Tjenesten behøver følgende tilganger:</p>
+            <ul class="sso-permissions">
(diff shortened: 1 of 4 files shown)


templates/oidc_provider/authorize.html
```diff
@@ -0,0 +1,42 @@
...
+            {% endfor %}
+            </ul>
+            <br />
+            <div id="sso-control-buttons">
+                <input type="submit" class="btn btn-large btn-success" name="allow" value="Godta" />
+                <input type="submit" class="btn btn-large btn-danger pull-right" value="Avslå" />
+            </div>
+        </form>
+    </div>
+    <p class="text-center"><small><i class="glyphicon glyphicon-lock"></i> Ekstern autentisering benytter OAuth 2.0 gjennom TLS</small></p>
```
duvholt: Copypaste? 😄

sklirg: Yes, is it not relevant anymore? Could probably reword it to say "using OAuth2 and OpenID".

sklirg: Also: I'm not sure I'd treat that as a blocking issue (and therefore a "request for change"), but rather a small refactoring task. This feature is interesting to get merged ASAP and such a small issue can be handled in a new issue (labeled easy!) instead IMO ;-)

duvholt: Didn't consider that OIDC is built on OAuth2 and therefore the view can also be used for OAuth2. 

Regarding "request for change": IMO it's better to fix quick issues like this early, otherwise they rarely get done. In this case it wasn't really an issue, but I thought it was supposed to say OpenID Connect instead of OAuth2.

PR #1873, opened 2017-08-07
Made it possible to arrange hobbygroups by priority

martinlunde:
Minor feature

## Code Checklist

## Description of changes
Added functionality such that hobbygroups are listed by priority


apps/hobbygroups/models.py
```diff
@@ -8,6 +8,7 @@ class Hobby(models.Model):
     description = models.TextField(max_length=250)
     image = models.ImageField(upload_to=settings.IMAGES_PATH)
     read_more_link = models.URLField(blank=True)
+    hobby_priority = models.IntegerField(default=0)
 
     def __str__(self):
-        return self.title
+        return (self.title + " [Priority: " + str(self.hobby_priority) + "]")
```
sklirg: Could use string formatting rather than concatenation for prettier code;
`return "{title} (Priority: {priority})".format(title=self.title, priority=self.hobby_priority)`

apps/hobbygroups/models.py
```diff
@@ -8,6 +8,7 @@ class Hobby(models.Model):
     description = models.TextField(max_length=250)
     image = models.ImageField(upload_to=settings.IMAGES_PATH)
     read_more_link = models.URLField(blank=True)
+    hobby_priority = models.IntegerField(default=0)
```
sklirg: `Hobby.hobby_priority` is a bit redundant. Could simply name it "priority" rather than hobby_priority, since we know we're on a Hobby object already.

PR #1889, opened 2017-09-03
Feature/alpine django

### LINKED ISSUE #1827: Look into using PyPDF2 to create thumbnails for Offline issues, removing the need for Imagemagick as a system dependency.

- [ ] Bug report
- [X] Feature request

TIL that PyPDF2 is pure python, so we can scrap imagemagick for PDF-to-png needed for the Offline thumbnails. That means less system deps, and possibly also reducing a bit of Docker base image headache.

Being free from the shackles of system-wide dependencies

System dep on imagemagick and bloating the base docker image

sklirg:

- [x] QA / Code Review

## Description of changes
Use alpine instead of debian as base for django image

### Build time and image sizes overview

| version | size | build time | including image pull |
| --- | --- | --- | --- |
| debian | 939 MB | 2 min 3 sec | false |
| debian | 939 MB | 2 min 27 sec | true |
| alpine | 499 MB | 2 min 9 sec | false |
| alpine | 499 MB | 2 min 12 sec | true |

#### python3-alpine - cached python:3-alpine
docker-compose build --no-cache django  1,37s user 0,15s system 1% cpu 2:09,05 total

| REPOSITORY | TAG | IMAGE ID | CREATED | SIZE |
| --- | --- | --- | --- | --- |
| onlineweb4_django | latest | ee940288165a | 24 seconds ago | 499MB |

#### python:3 - cached python:3
docker-compose build --no-cache django  1,33s user 0,18s system 1% cpu 2:03,65 total

| REPOSITORY | TAG | IMAGE ID | CREATED | SIZE |
| --- | --- | --- | --- | --- | 
| onlineweb4_django | latest | cd17fe69de66 | 5 seconds ago | 939MB |

#### python:3-alpine - pull
docker-compose build --no-cache django  1,56s user 0,22s system 1% cpu 2:12,83 total

#### python:3 - pull
docker-compose build --no-cache django  1,50s user 0,22s system 1% cpu 2:27,84 total

### Conclusion

While the build times are quite similar, the image sizes are reduced to half.

They can further be reduced if we find ways to remove dependencies to e.g. `git` (and probably `base-build`)  (by not installing dependencies using `git+https://` in `requirements.txt`) and if we manage to find a fix for #1827.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -2,6 +2,8 @@
 import os
 import sys
 
+import threading; threading.stack_size(4*80*1024)
+
 if __name__ == "__main__":
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "onlineweb4.settings")
 
--- a/Dockerfile.django
+++ b/Dockerfile.django
@@ -1,4 +1,4 @@
-FROM python:3
+FROM python:3-alpine
 MAINTAINER dotkom
 
 ENV APP_DIR=/srv/app
@@ -6,10 +6,11 @@ ENV APP_DIR=/srv/app
 RUN mkdir -p $APP_DIR
 WORKDIR $APP_DIR
 
-RUN apt-get update
-RUN apt-get install -y --no-install-recommends \
-    libjpeg-dev \
-    ghostscript
+RUN apk add --no-cache \
+    build-base \
+    git \
+    imagemagick zlib-dev jpeg-dev \
+    postgresql-dev
 
 COPY requirements.txt $APP_DIR/requirements.txt
 RUN pip install -r $APP_DIR/requirements.txt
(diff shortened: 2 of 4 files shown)


sklirg:
Forgot to fix testbase image too, will look at some other time.

manage.py
```diff
@@ -2,6 +2,8 @@
 import os
 import sys
 
+import threading; threading.stack_size(4*80*1024)
```
duvholt: A comment on why this is need would be nice

sklirg: I don't know _why_ it's needed 🙈.

Random stack overflow posts or github issues resolved it by manually setting the stack size. We should do some research on what the implications of this is.

sklirg:
I think I'll abandon these changes. If someone wants to re-raise the changes, feel free.

PR #1931, opened 2017-10-18
Add default title for base template, remove unique SSO base template

### LINKED ISSUE #1857: SSO base template should extend the base template

- [x] Bug report

The template for the SSO app should extend the base template so we don't have to maintain multiple base templates.

It's a custom template, which looks copy/pasted from the base template.

sardred:
## Code Checklist

## Description of changes

- [x] Added default title to base template
- [x] Removed myth's phantom base template, instead extending base directly from sso/authorize (the only template for sso)

This should result in an extremely minor change in design, namely the features from base that were (supposedly) arbitrarily removed from the sso/base.

Why? Because it fixes #1857


sklirg:
Deleting the sso/index.html template removes the template used for OpenID Connect as well. Could you fix that too?

https://github.com/dotkom/onlineweb4/blob/develop/templates/oidc_provider/authorize.html

sardred:
I looked at fixing the OpenID connect as well, only reason it hasn't been pushed is because I couldn't figure out how to test it locally, so I'm not sure the fix works. It should, but to be safe I'm delaying it. 
This is a really low priority PR regardless, so it'll wait until after my crunchtime period this semester.

sklirg:
You could create a new branch with those changes and open a Pull Request towards this branch as well. You would have to revert the commit regarding deleting the sso_base template (in this branch), though, since SSO should continue using that template until a fix is in place.

If so: 
* Revert change that deletes base_sso template
* Push change
* Create a new branch
* Fix template inheritance
* Delete base_sso template
* Open a PR towards this branch (`refactor/sso_base`)

If this PR is merged before that one, we could change the _base_ of that pull request to be against develop rather than this branch, since it has been merged.

However, as of now, until we can confirm that the SSO base template is fixed as well, we cannot merge this PR if we want to use OIDC.

sardred:
oh fuck.

PR #1948, opened 2017-11-11
Parallelize drone CI tests

sklirg:
## Description of changes

This way you'll potentially get feedback a bit quicker since tests run in parallell.

Previously you could have two errors in different test runs (e.g. one lint error and one test error). The tests would exit after the first non-zero error code, which hid the fact that you failed tests. Only when you fixed the linting problem the failing test would appear. This is a silly way of doing things.

Also, it should be easier to see what category the test failure is in, since each build step has a name which tells you what's there (linting, testing, ...)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -14,7 +14,7 @@ pipeline:
     image: registry.online.ntnu.no/dotkom/onlineweb4-testbase
     when:
       event: [push, pull_request, tag, deployment]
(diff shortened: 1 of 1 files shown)


sklirg:
I didn't change that line you p$/&e of sh¤&!!!!!!!!!!!!!!!!!!!!!

Issue #1950, opened 2017-11-14
500 server error on admin documentation 


thommath:
## What kind of an issue is this?

- [x] Bug report
- [ ] Feature request

## What is the expected behaviour?
Either link to some dokumentasjon that exists or remove the button (located in /admin upper right corner) if not used

## What is the current behaviour?
When logged in as admin this returns 500 internal server error. 
https://online.ntnu.no/admin/doc/

## How do you reproduce this problem? 
- open https://online.ntnu.no/admin/doc/
- if not logged in, log in

PR #1951, opened 2017-11-15
Solve issue #1950, internal server error when accessing docs

martinlunde:
## Description of changes
Adds docutils to dependencies and django.contrib.admindocs to installed apps


martinlunde:
#1950

PR #1974, opened 2017-12-02
Secret santa

MrHanSan:

- [ ] QA / Code Review

## Code Checklist

## Description of changes
Made secret santa pair generator function run at a more predictable speed. Also changed the mail templates to give a more accurate description.


scripts/management/commands/secret_santa.py
```diff
@@ -24,31 +24,41 @@ def handle(self, *args, **options):
...
     # Checks the index of the original list of users against the randomly
     # generated number. If index and random number is the same, then someone
     # got themselves as secret Santa.
     @staticmethod
-    def check_random_list(random):
-        for user_index, user2_index in enumerate(random):
-            if (user_index == user2_index):
-                return False
-        return True
+    def check_random_list(random_list):
```
sklirg: If this function makes sure the map follows the set constraints rather than just checking that it doesn't violate the constraints, the name should probably be changed from `check` to `fix` or something along those lines.

templates/secret_santa/committee_mail.txt
```diff
@@ -1,2 +1,4 @@
 Secret santa e-post har blitt sendt ut.
-Det er lagret en kopi av alle til-fra par om det blir behov for å sjekke dette.
\ No newline at end of file
+
+Det er lagret en liste med alle til-fra par.
+Ta kontakt med DotKom om om det blir behov for å sjekke denne.
```
sardred: It's not DotKom, nor Dotkom, it's just dotkom.

scripts/management/commands/secret_santa.py
```diff
@@ -24,31 +24,41 @@ def handle(self, *args, **options):
...
+        self.mail_to_committee()
 
     # Function for generating a list with random numbers corresponding to
     # the indexes of the list of users attending the event.
     @staticmethod
     def generate_random_list(users):
         to_from = random.sample(range(len(users)), len(users))
-        while(not Command.check_random_list(to_from)):
-            to_from = random.sample(range(len(users)), len(users))
+        Command.check_random_list(to_from)
```
sklirg: The `check_random_list` function mutates the to_from list and returns the updates list. While the current implementation _works_, it's better practice to assign the returned value from the function to a variable. This makes the code less magic. :-)

You could also copy (`to_from[:]`) the list to make sure you're not mutating the list, but rather working on a copy of it.

If you did that here, you would get different results, since you don't store the results of the function but rather expect it to mutate the referenced list.

scripts/management/commands/secret_santa.py
```diff
@@ -24,31 +24,41 @@ def handle(self, *args, **options):
...
-                return False
-        return True
+    def check_random_list(random_list):
+        for user_index, user2_index in enumerate(random_list):
+            if user_index == user2_index:
+                Command.swap_pair(random_list, user_index)
+        return random_list
+
+    @staticmethod
+    def swap_pair(random_list, user_index):
```
sklirg: This function mutates the list it receives and returns nothing. That's quite magic, as commented [here](https://github.com/dotkom/onlineweb4/pull/1974/files#r154499798). Might look into making it less magic and actually copy the list and return the result.

scripts/management/commands/secret_santa.py
```diff
@@ -24,31 +24,41 @@ def handle(self, *args, **options):
...
+        user_swap_index = random.choice(random_list)
+        if user_swap_index == user_index:
+            if user_swap_index != len(random_list)-1:
+                user_swap_index += 1
+            else:
+                user_swap_index -= 1
+        random_list[user_index], random_list[user_swap_index] = \
+            random_list[user_swap_index], random_list[user_index]
+        Command.check_random_list(random_list)
```
sklirg: Assign the result here to the existing list rather than expecting it to mutate the object you send in.

templates/secret_santa/committee_mail.txt
sklirg: #altvarbedrefør

MrHanSan:
Made the pair generator less magic, by creating copies of the list, and manipulating and returning the copy.

The script should also run faster, since it no longer have to re-run the entire process of checking the validity of the to-from list after every edit.

PR #1978, opened 2017-12-07
URL Integration tests for some event views

duvholt:
Tests for viewing event(includes group restrictions, extras and payment), choosing extras, attending and unattending. 

Also contains some minor fixes that I found while adding tests:

- Custom captcha errors weren't used
- Error checks done while choosing extras weren't working

### DIFF
--- a/apps/events/tests/utils.py
+++ b/apps/events/tests/utils.py
@@ -32,11 +32,13 @@ def attend_user_to_event(event, user):
     )
 
 
-def pay_for_event(event, user):
+def pay_for_event(event, user, *args, **kwargs):
     return G(
         PaymentRelation,
         payment=event.attendance_event.payment(),
-        user=user
+        user=user,
+        *args,
+        **kwargs
     )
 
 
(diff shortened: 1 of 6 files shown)


apps/events/tests/utils.py
```diff
@@ -1,17 +1,55 @@
...
     G(AttendanceEvent, event=event)
     return event
 
 
+def generate_payment(event):
+    return G(
+        Payment,
+        object_id=event.id,
+        content_type=ContentType.objects.get_for_model(AttendanceEvent)
+    )
```
sklirg: Some crazy formatting here, but whatever floats your ⛵️

duvholt: To be honest I like it better than what autopep8 produces.
`object_id=event.id` could've been put on the line above, but I think it's clearer when each kwarg is on its own line. 

I especially don't like how arguments should line up with the line above.  http://docs.python-requests.org/en/latest/dev/contributing/#kenneth-reitz-s-code-style

Anyway doesn't really matter and I don't think we have any consist styling when it comes to this anyway.

sklirg: Nothing to keep this from being merged at all, no.

We could look into using [yapf](https://github.com/google/yapf) if we want to enforce some rules regarding this.

PR #1979, opened 2017-12-07
Check if user is not already an attendee before attending

### LINKED ISSUE #1764: Check if user is attending before trying to attend user

duvholt:
Fixes #1764

### DIFF
--- a/apps/events/models.py
+++ b/apps/events/models.py
@@ -598,6 +598,7 @@ class AttendanceEvent(models.Model):
         """
         Checks if a user can attend a specific event
         This method checks for:
+            Already attended
             Waitlist
             Room on event
             Rules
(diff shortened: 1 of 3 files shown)


apps/events/models.py
```diff
@@ -617,6 +618,12 @@ def is_eligible_for_signup(self, user):
 
         response = {'status': False, 'message': '', 'status_code': None}
 
+        # User is already an attendee
+        if self.attendees.filter(user=user).count() > 0:
```
sklirg: I like to use `Model.object.get(**filters)` rather than doing an if/else for `.count()`, but that might just be me.

duvholt: IMO it makes more sense to check if something exists instead of trying to retrieve something only to catch its exception. 
I know it's more common in Python to use exceptions for control flow, but I don't think it fits here. 

Anyway out of curiosity I searched to see if there were anyway preferred way to check if an object exists, and Django apparantly has `queryset.exists()`. That's pretty much perfect for this use case.

PR #1983, opened 2017-12-21
Add tox to run tests

sklirg:

- [x] QA

## Code Checklist

## Description of changes

Simplify running tests by adding a tox config to run them with.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -43,17 +43,6 @@ pipeline:
     commands:
       - npm run lint-less
 
-  python-tests:
-    image: registry.online.ntnu.no/dotkom/onlineweb4-testbase
-    secrets: [codecov_token]
-    when:
-      event: push
-    group: testing
-    commands:
-      - tox -e tests
-      - pip install codecov --quiet
-      - codecov
-
   python-lint:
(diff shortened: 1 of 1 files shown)


sklirg:
We had tox a year or so ago, but we removed it when we stopped using it after changing CI systems.

Quote [Tox homepage](https://tox.readthedocs.io/en/latest/):
> vision: standardize testing in Python
> tox aims to automate and standardize testing in Python. It is part of a larger vision of easing the packaging, testing and release process of Python software.

The point of adding tox again is to simplify running tests for ow4 (and other python projects), by simply installing tox and running tox in a python project. It will handle installing the required dependencies by itself, so you don't need to do that by yourself (or run it in the same docker-compose setup as the ow4 dev experience).

"Manually" running the tests requires you to check what command is executed in the test environment and run that. By adding tox we can create this command in tox and people can run the tox env they want to, rather than install and copy a command (`tox -e isort` vs check what requirements are required, install requirements, find command, execute command (worst case)).
Currently we have a makefile which simplifies running tests, but a makefile doesn't work out of the box on e.g. windows.

Hope that shed some light on the propsed changes.

PR #1986, opened 2017-12-23
Sphinx documentation

duvholt:
Mostly boilerplate documentation for API references (models, views etc).

As a start I wrote some documentation for parts of authentication and payment. 
This should hopefully make it easier to contribute with more documentation. 

Sphinx is what Python, Django and a bunch of other projects use to write their documentation. 
The main use is generating HTML pages we can host on something like docs.online.ntnu.no. 

Before this can be merged we need a better way to import other settings files. 
Sphinx crashes unless this line is removed:
https://github.com/dotkom/onlineweb4/blob/03a42d7115a31daf02291c074163cadfd062c52c/onlineweb4/settings/base.py#L539


apps/payment/models.py
```diff
@@ -33,26 +33,46 @@ class Payment(models.Model):
     )
 
     content_type = models.ForeignKey(ContentType)
+    """Which model the payment is created for. For attendance events this should be event."""
     object_id = models.PositiveIntegerField()
+    """
+    Object id for the model chosen in content_type.
+    For events this should be the event id and not attendance event id.
```
sklirg: The attendance event id will always be equal to the event id, since the primary key of an attendance event is an event.
https://github.com/dotkom/onlineweb4/blob/develop/apps/events/models.py#L399

apps/payment/models.py
```diff
@@ -33,26 +33,46 @@ class Payment(models.Model):
...
         default="arrkom"
     )
+    """Which Stripe key to use for payments"""
 
     payment_type = models.SmallIntegerField(_('type'), choices=TYPE_CHOICES)
+    """
+        Which payment type to use for payments.
+        Can be one of the following:
+
+        - Immidiate
```
sklirg: Should probably be "Immediate".

apps/payment/models.py
```diff
@@ -33,26 +33,46 @@ class Payment(models.Model):
...
+        - Delay. See :attr:`delay`
+    """
 
-    # Optional fields depending on payment type
     deadline = models.DateTimeField(_("frist"), blank=True, null=True)
+    """Shared deadline for all payments"""
     active = models.BooleanField(default=True)
+    """Is payment activated"""
     delay = models.SmallIntegerField(_('utsettelse'), blank=True, null=True, default=2)
+    """Number of days before user has to pay"""
```
sklirg: Could probably add "after attending", as in "Number of days after user attended which they have to pay."

docs/conf.py
```diff
@@ -0,0 +1,189 @@
...
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+import django
+import os
+import sys
+from django.conf import settings
+
+sys.path.insert(0, os.path.abspath('../'))
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "onlineweb4.settings")
+#settings.configure()
```
sklirg: Why is this commented out?

docs/payment/index.rst
```diff
@@ -0,0 +1,98 @@
...
+
+To enable the payment functionality on an Event a :class:`~apps.payment.models.Payment` object should be created. 
+See :class:`~apps.payment.models.Payment` for documentation of the attributes required.
+
+Attendance
+**********
+
+Depending on which :attr:`~apps.payment.models.Payment.payment_type` was selected three different things can happen.
+
+- **Immidiate**: Nothing? (Should be looked into)
```
sklirg: Should probably also be "Immediate".

docs/payment/index.rst
```diff
@@ -0,0 +1,98 @@
...
+If a user has already paid tey are not allowed to unattend events without being refunded first.
+Otherwise any payment delays are just deleted if payment type is set to delay.
+
+
+Waitlist bump
+*************
+
+Depending on which :attr:`~apps.payment.models.Payment.payment_type` was selected three different things can happen.
+
+- **Immidiate**: :class:`~apps.payment.models.PaymentDelay` is created with a delay of two days
```
sklirg: Should probably also be "Immediate".

apps/payment/models.py
duvholt: Ah, that makes sense.

docs/conf.py
duvholt: Tried a bunch of stuff to make Sphinx works with Django and forgot to remove some of it.

duvholt:
I looked into splitting the docs folder, but because of how Sphinx works this doesn't seem possible unless we either move the main docs folder to the root folder or create include files for each .rst file. 
Neither of these solutions seems good to me. 

Fixed the other comments.

PR #1991, opened 2017-12-26
Comply with stylelint-config-standard

sardred:

- [x] One of those big old ones that inflate your "contributors" stats

## Code Checklist

Stylelint considers [no-descending-specificity](https://stylelint.io/user-guide/rules/no-descending-specificity/) to be a mandatory error rule, as selectors with varying specificity can overlap, meaning the order of the styles ends up deciding what rule applies. While the entire rule exists to prevent edge cases, we _might_ have abused this edge case in our style, meaning that fixing the rule may break some styling. 

## Description of changes

Ran an automatic `stylelint --fix` through `yarn lint-less`, which surprisingly doesn't cover a whole lot of rules you'd think would be easily automated. A lot of manual work to fix the remaining rules, later realizing that no-descending-specificity could break some styling (read above). Can't easily test that locally, but I did do my best in terms of double-checking that selectors are the same, just different places (or different level of nesting).

### DIFF
--- a/assets/core/less/navbar.less
+++ b/assets/core/less/navbar.less
@@ -76,10 +76,6 @@
   border-top: 5px solid rgba(255, 255, 255, 1);
 }
 
-.mn-nav li:hover a {
-  text-decoration: none;
-}
-
 .mn-nav li a {
   display: block;
   line-height: @navbar-height;
@@ -87,6 +83,10 @@
   padding: 0 13px;
   text-transform: uppercase;
   color: #fff;
+}
+
+.mn-nav li:hover a {
+  text-decoration: none;
 }
 
 .mn-nav li:first-child img {
--- a/assets/core/less/navbar.less
+++ b/assets/core/less/navbar.less
@@ -8,13 +8,13 @@
 
 @navbar-height:                    57px;
 
-@navbar-default-color:             #FFF;
+@navbar-default-color:             #fff;
 @navbar-default-bg:                #004b80;
 
-@navbar-default-link-color:                #FFF;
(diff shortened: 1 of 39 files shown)


assets/core/less/navbar.less
```diff
@@ -1,437 +1,439 @@
...
-@state-warning-text:             #c09853;
-@state-warning-bg:               #fcf8e3;
-@state-warning-border:           darken(spin(@state-warning-bg, -10), 3%);
+.mn-nav li > .online-logo-link {
+  padding: 0 13px 0 0;
+}
 
-@state-danger-text:              #b94a48;
-@state-danger-bg:                #f2dede;
-@state-danger-border:            darken(spin(@state-danger-bg, -10), 3%);
```
duvholt: I don't think these variables are actually unused. 
They overwrite Bootstrap variables and are probably not used by us directly. 
Odd place to overwrite this, but that's another issue. 

Do you know if stylelint takes into account less files imported from node_modules too?

sardred: We had a quick chat on Slack, but for the sake of clarity:
I'm fairly sure the variables are unused, at least I can't spot a difference to the site when removing them. You're right on them overwriting Bootstrap, which is why I didn't spot their usage. Stylelint doesn't check for unused variables, that was just my own doing, which was partly why I put it in its own commit in case I was wrong. 
I'm somewhat :tm: sure that we can just remove them, but since I want to do a bigger sweep of unused styling in the future, I just moved them into bootstrap-variables.less.

PR #1995, opened 2017-12-28
Hotfix to offline-archive button looking mildly different from the ot…

sardred:
…hers
image
It fixes that garbage.


duvholt:
Good branch name.

PR #1996, opened 2017-12-28
Fix #1919. Use python nonzero method rather than 'is not None'

### LINKED ISSUE #1919: Attendance API tries to query attendees by RFID value "None"
Sometimes the regme application will (probably?) send the "rfid" payload as null which queries the database for `event=event_id` and `user__rfid=None`. This might result in a "MultipleObjectsReturned" since there may be more than one attended user which haven't set up their RFID.

Can be reproduced;
* Create attendance event
* Attend at least two people who does not have set their RFID
* Try to verify attendance  of the users by POSTing `rfid` payload as null

---

https://sentry.io/dotkom/onlineweb-4/issues/353044724/

```
MultipleObjectsReturned: get() returned more than one Attendee -- it returned 2!
(7 additional frame(s) were not displayed)
...
  File "rest_framework/views.py", line 483, in dispatch
    response = self.handle_exception(exc)
  File "rest_framework/views.py", line 443, in handle_exception
    self.raise_uncaught_exception(exc)
  File "rest_framework/views.py", line 480, in dispatch
    response = handler(request, *args, **kwargs)
  File "apps/events/views.py", line 364, in post
    attendee = Attendee.objects.get(event=event, user__rfid=rfid)

MultipleObjectsReturned: get() returned more than one Attendee -- it returned 2!
```

sklirg:

- [x] QA / Code Review

## Code Checklist

    - Depending on the below todos.

    - [x] The changes might require changes in third party applications (dotkom/regme)

## Description of changes

Use python nonzero method rather than checking if the input actually is None. It might actually be empty string, which is not None. [Read more](https://stackoverflow.com/questions/7816363/if-a-vs-if-a-is-not-none)

```
>>> a = ''

# Check that the variable is not none, so that we can proceed knowing that the input actually is not None
>>> bool(a is not None)
True

# But what if a is '' (empty string) instead? That should probably also halt execution.
>>> bool(a)
False
```

### DIFF
--- a/apps/events/views.py
+++ b/apps/events/views.py
@@ -402,6 +402,10 @@ class AttendViewSet(views.APIView):
         username = request.data.get('username')
         waitlist_approved = request.data.get('approved')
 
+        if not (username or rfid):
+            return Response({'message': 'Mangler både RFID og brukernavn. Vennligst prøv igjen.',
+                             'attend_status': 41}, status=status.HTTP_400_BAD_REQUEST)
+
         # If attendee has typed in username to bind a new card to their user
         if username is not None and rfid is not None:
             try:
@@ -443,11 +447,6 @@ class AttendViewSet(views.APIView):
                 return Response({'message': 'Brukernavnet finnes ikke. Husk at det er et online.ntnu.no brukernavn! '
                                             '(Prøv igjen, eller scan nytt kort for å avbryte.)', 'attend_status': 50},
                                 status=status.HTTP_400_BAD_REQUEST)
-
-            # If RFID is empty string or the likes.
-            elif not rfid:
-                return Response({'message': 'RFID-en er ugyldig. Ble kortet scannet korrekt? Prøv igjen.',
-                                 'attend_status': 41}, status=status.HTTP_400_BAD_REQUEST)
 
             # If attendee tried to attend by card, but card isn't tied to a user
             else:
--- a/apps/events/views.py
+++ b/apps/events/views.py
@@ -414,7 +414,7 @@ class AttendViewSet(views.APIView):
                                 status=status.HTTP_400_BAD_REQUEST)
         try:
             # If attendee is trying to attend by username
-            if rfid is None:
+            if not rfid:
                 attendee = Attendee.objects.get(event=event, user__username=username)
             else:
                 attendee = Attendee.objects.get(event=event, user__rfid=rfid)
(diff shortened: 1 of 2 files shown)


sklirg:
Okay so I'm going away for a couple of days. This view probably warrants a full refactor since it's bordering being too complex. If anyone wants a go at it, feel free to go, otherwise it might not be done before next year. (ha ha that's funny, right?)

apps/events/views.py
```diff
@@ -402,20 +422,12 @@ def post(self, request, format=None):
...
-                user.save()
-            except User.DoesNotExist:
-                return Response({'message': 'Brukernavnet finnes ikke. Husk at det er et online.ntnu.no brukernavn! '
-                                            '(Prøv igjen, eller scan nytt kort for å avbryte.)', 'attend_status': 50},
-                                status=status.HTTP_400_BAD_REQUEST)
+        error = validate_attend_params(rfid, username)
+        if 'message' in error and 'attend_status' in error:
+            return Response({'message': error.get('message'), 'attend_status': error.get('attend_status')},
+                            status=status.HTTP_400_BAD_REQUEST
+                            )
```
duvholt: Formatting 🤢

apps/events/views.py
```diff
@@ -390,6 +390,26 @@ class CompanyEventViewSet(viewsets.GenericViewSet, mixins.RetrieveModelMixin, mi
     permission_classes = (AllowAny,)
 
 
+def validate_attend_params(rfid, username):
```
duvholt: Wouldn't this work better as a method on AttendViewSet?

apps/events/views.py
sklirg: What's wrong?

apps/events/views.py
sklirg: There's nothing wrong by adding it to the viewset, but I don't see any difference in doing so. It could potentially in theory be reused other places if needed (e.g. other (future) regme-endpoints, nibble, ...)

apps/events/views.py
duvholt: It's probably fine by pep8 standards and just me being weird.
It was just the combination of a bunch of spaces to match indention level of the line above and adding the ending parenthesis on a new line with the same indention level. 

In cases like this I prefer 
```python
return Response(
    {'message': error.get('message'), 'attend_status': error.get('attend_status')},
    status=status.HTTP_400_BAD_REQUEST
)
```

or (but this is probably not allowed by pep8 idk)

```python
return Response({'message': error.get('message'), 'attend_status': error.get('attend_status')},
    status=status.HTTP_400_BAD_REQUEST)
```

Either way feel free to ignore the comment.

apps/events/views.py
duvholt: Sounds like YAGNI, but meh. Ække så nøye.

PR #1998, opened 2017-12-28
Use yarn in Drone

duvholt:
Since yarn can use our lockfile this avoids the 1min setup npm does, which brings our total build time down to 5min.


duvholt:
Yeah, that was done because I was lazy and wanted to see how fast the build could potentially be.
I think it's likely that I'll merge the other PR first anyway.

PR #2002, opened 2018-01-02
Started refactoring event model methods

### LINKED ISSUE #1183: Payment - Users did not get paymentdelay when bumped from waitlist
Think this happened when reserved spots was released.

duvholt:
Most of the changes are just cleaning up old code, but I also hid some bug fixes in there. 😇

- Check for waitlist bump on AttendanceEvent.save instead of only when saving event in admin
- Take into account reserved seats when checking for waitlist bump. Fixes #1183
- `is_attendance_event` was not called in `feedback_users`.

**I added unit tests for most of the changes, but I haven't done any major manual testing.**

If someone want to bikeshed a better name for `attending_attendees_qs` feel free. It was called `attendee_qs`, but imo that's a shitty name when `attendees` is everyone including waitlist. 
Changing `attendees` is not as straight forward as the model is called Attendee and would require a migration.

* [x] Do manual testing


sklirg:
> Changing attendees is not as straight forward as the model is called Attendee and would require a migration.

Yes, but not necessarily in a major way. It's possible to rename the model but keep the old database table (which then doesn't require executing any SQL) by using the Model Meta class option `db_table`, which can be set to `events_attendee` through a migration like so (it's actually a NoOp, since that's the actual name of the database table);

```
operations = [
    migrations.AlterModelTable(
         name='attendee',
        table='events_attendee',
    ),
]
```

After which it can be renamed and doesn't require altering the table.

https://docs.djangoproject.com/en/2.0/ref/models/options/#db-table

sklirg:
To answer the question though;

> If someone want to bikeshed a better name for `attending_attendees_qs` feel free. It was called `attendee_qs`, but imo that's a shitty name when attendees is everyone including waitlist.

What about

`AttendanceEvent.attendees` - all attendees / registrants / participants, attending and on waitlist
`AttendanceEvent.attending` - actually attending attendees (before waitlist)
`AttendanceEvent.waitlist` - attendees on waitlist

The latter two could be prefixed with `attendees_` if needed.

sklirg:
Another idea might be to deprecate `.attendee_qs` and rather create some new methods, like `AttendanceEvent.registrants` or `.participants` or whatnot, and `AttendanceEvent.waitlist`; and use them instead, and finally remove `.attendee_qs` in the future.

This should probably warrant a rename of the Attendee-model.

apps/events/models.py
```diff
@@ -419,52 +403,43 @@ class AttendanceEvent(models.Model):
...
-    def attendees_qs(self):
+    def attending_attendees_qs(self):
         """ Queryset with all attendees not on waiting list """
-        return self.attendees.all()[:self.max_capacity - self.number_of_reserved_seats]
+        return self.attendees.all()[:self.number_of_attendee_seats]
 
     def not_attended(self):
         """ Queryset with all attendees not attended """
         # .filter does apperantly not work on sliced querysets
         # return self.attendees_qs.filter(attended=False)
```
sklirg: Could probably remove these comments as well, I don't think they're relevant any more.

apps/events/models.py
```diff
@@ -761,7 +720,7 @@ def is_attendee(self, user):
         return self.attendees.filter(user=user)
 
     def is_on_waitlist(self, user):
```
sklirg: If you're in here refactoring, could this be renamed to `attendees_on_waitlist`? This method sounds like one you can use to check if a user is on the waitlist, like `is_on_waitlist(sklirg) : bool`.

apps/events/models.py
```diff
@@ -793,8 +754,30 @@ def get_payment(self):
     def __str__(self):
         return self.event.title
 
+    def check_for_waitlist_changes(self):
```
sklirg: I'm a bit conflicted about this method.

I understand what it tries to do and I like it, but I think it should require one (or two (or one and one optional)) event(s) as arguments, so that it can be used outside of the `save`-method, which it now is tightly coupled to.

Like `def check_for_waitlist_changes(self, old_event)`, or even as a classmethod with `cls, updated_event, old_event` if that's your fancy.

Also, when I'm here: I don't like a function called `check_something` that actually does an action; I'd expect it to return a value I can use myself to do an action if the check is truthy (or falsy, for that matter), not send emails to some people. ;-)

apps/events/models.py
```diff
@@ -858,6 +841,9 @@ def delete(self):
             # Do nothing
             False
 
+        if not self.is_on_waitlist():
+            self.event.notify_waiting_list()
```
sklirg: I'm reading this as "if noone is on waitlist; notify waitlist". Is that correct? 🤔

apps/events/pdf_generator.py
```diff
@@ -21,7 +21,7 @@ class EventPDF(object):
 
     def __init__(self, event):
         self.event = event
-        attendee_qs = event.attendance_event.attendees_qs
+        attendee_qs = event.attendance_event.attending_attendees_qs
```
sklirg: Super nitpick: If you refactor what field you proxy, you should probably also rename the variable that proxies it. :p

apps/events/models.py
duvholt: But that's exactly what it does? It just does the filtering in Python because we can't use filter on spliced querysets.

apps/events/models.py
duvholt: I get what you mean, but it's not really coupled to the save method. 
It would however not make any sense to use this without calling .save after though, but I can't think of any other ways to do this check without modifying the fields. 
Using a classmethod seems silly to me as you then have to manually call save on one of the objects after. 

I agree on the method name though. Any suggestions?

apps/events/models.py
duvholt: No, it should be read as `if not attendee is on waitlist`.
Some context is lost in the diff as this is really a method on the Attendee class. 
More specifically Attendee.delete.

apps/events/models.py
sklirg: Wow, I must be blind.

apps/events/models.py
sklirg: True, but it could be used to inform people what is going to happen, e.g. if someone increases the capacity of the event, it could inform the user that "This action is going to give 5 attendees currently on waitlist space on this event. Confirm this?".

Classmethod was a stretch, disregard that.

What I meant was that `save()` could call this method, and if it's truthy it can do `notify_waiting_list()`:

```
def save(…):
    if self.check_for_waitlist_changes():
        self.notify_waiting_list()
    ...
    super().save()
```

was more along the lines of what I thought.

apps/events/models.py
sklirg: Ah, right, that's where the confusion arised.

apps/events/models.py
duvholt: Yeah, that makes sense. I could rewrite it to return the number of attendees bumped from the waitlist and call notify_waiting_list with that number if > 0.

sklirg: :+1:

duvholt:
I decided to go for deprecating waitlist_qs etc., but that'll be handled in another PR.

PR #2010, opened 2018-01-12
Requires io/temp

sklirg:

- [x] Requirements bump

## Code Checklist
- [ ] The code is ready to be merged ⚠️ 

## Description of changes

This is #1977 but without the Django 2.0 bump.

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -8,6 +8,6 @@ isort==4.2.15
 freezegun==0.3.9
 
 # py.test
-pytest==3.3.0
+pytest==3.3.1
 pytest-cov==2.5.1
 pytest-django==3.1.2
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -8,6 +8,6 @@ isort==4.2.15
 freezegun==0.3.9
 
 # py.test
-pytest==3.3.1
+pytest==3.3.2
 pytest-cov==2.5.1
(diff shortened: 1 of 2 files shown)


sklirg:
I looked up the changelogs and they look fine for our use.

Django-wiki only tests against django-mptt 0.8.6, while requires-io wants us to bump to 0.9. Do you know if this is safe for django-wiki, @duvholt?

Is there a reason why we install and pin django-mptt ourselves, could we not depend on django-wiki requirements?

duvholt:
>Django-wiki only tests against django-mptt 0.8.6, while requires-io wants us to bump to 0.9. Do you know if this is safe for django-wiki

Django-wiki specifies <0.9 and has likely not been tested with 0.9. 

>Is there a reason why we install and pin django-mptt ourselves, could we not depend on django-wiki requirements?

I think it was done like this to force upgrading packages to something that is Django 1.x compatible.  
We don't use django-mptt directly, so I think it's safe to remove mptt and sekizai. 
Either that or add >=version.  

Could you squash all the requires.io dependency update commits btw? Really wish requires.io included some meaningful info in their commit messages.

sklirg:
Drone/tox is not actually testing the updated requirements, because of this bug: https://github.com/tox-dev/tox/issues/149

I tested locally and all tests pass, but we should look into how we can fix this issue.

PR #2014, opened 2018-01-26
Transaction receipts

martinlunde:
Make all transactions use an universal receipt template.

- [x] QA / Code Review

## Code Checklist

## Description of changes
Instead of using different functions for sending transaction confirmations, we will be able to create "receipt" objects which can be sent to users with a standard _send_receipt() function. This function uses a template and ensures that every transaction mail will have the same setup.


apps/payment/models.py
```diff
@@ -298,9 +297,32 @@ def save(self, *args, **kwargs):
...
 
     class Meta:
         ordering = ['-datetime']
         verbose_name = _('transaksjon')
         verbose_name_plural = _('transaksjoner')
+
+
+class PaymentReceipt(models.Model):
+    """Transaction receipt"""
+    receipt_id = models.CharField(max_length=128, default=None)
```
sklirg: Maybe use UUIDField for this? Generates unique IDs for us. Can also be used as primary key (`primary_key=True`).

sklirg: Maybe use UUIDField for this? Generates unique IDs for us. Can also be used as primary key (`primary_key=True`).

apps/payment/models.py
```diff
@@ -298,9 +297,32 @@ def save(self, *args, **kwargs):
...
+    to_mail = models.EmailField()
+    from_mail = models.EmailField()
+    subject = models.TextField()
+    description = models.TextField()
+    transaction_date = models.DateTimeField()
```
sklirg: If you connect the receipt to a specific payment you could get a lot of these fields for free. Event payments have user (from which you can get to_email), subject (event title), description (what you purchased) [or the other way around], as well as the transaction date.

apps/payment/models.py
```diff
@@ -298,9 +297,32 @@ def save(self, *args, **kwargs):
...
+
+    def save(self, *args, **kwargs):
+        if not self.receipt_id:
+            self.receipt_id = str(uuid.uuid4())
+        super().save(*args, **kwargs)
+
+
+class ReceiptItem(models.Model):
```
sklirg: If you connect PaymentReceipt to a Payment, the Payment will likely have its own (sub)items attached, therefore making this model redundant.

apps/payment/models.py
sklirg: (And when I come to think of it, calling save() on a payment could actually automagically create a PaymentReceipt, which is automagically sent to the user. Neat, huh?)

apps/payment/views.py
```diff
@@ -52,7 +50,23 @@ def payment(request):
...
+                    receipt.save()
+
+                    item = ReceiptItem(
+                        receipt=receipt,
+                        name=payment_relation.payment.description(),
+                        price=int(payment_relation.payment_price.price)
+                    )
+                    item.save()
+
+                    _send_receipt(receipt)
```
sklirg: If adding support for automagically sending PaymentReceipts as mentioned in models, this might be redundant ;-)

apps/payment/views.py
```diff
@@ -241,9 +251,27 @@ def saldo(request):
...
                 )
 
-                PaymentTransaction.objects.create(user=request.user, amount=amount, used_stripe=True)
+                payment_transaction = PaymentTransaction.objects.create(user=request.user, amount=amount,
+                                                                        used_stripe=True)
+
+                receipt = PaymentReceipt(
+                    to_mail=payment_transaction.user.email,
+                    from_mail=settings.EMAIL_TRIKOM,
+                    subject="[kvittering] saldo inskudd",
```
sklirg: Could probably capitalize `[Kvittering]`. Also, it's gramatically correct to write "Saldoinnskudd". 

While you're at it, it could be even more descriptive: "Saldoinnskudd på online.ntnu.no" or something the like.

apps/payment/models.py
```diff
@@ -298,9 +302,110 @@ def save(self, *args, **kwargs):
...
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    """Which model the receipt is created for"""
+    object_id = models.PositiveIntegerField(null=True)
+    """Object id for the model chosen in content_type"""
+    content_object = GenericForeignKey()
+    """Helper attribute which points to the object select in object_id"""
+
+    def save(self, *args, **kwargs):
+        # Saldo
+        if self._is_type(PaymentTransaction):
```
sklirg: The logic in these if/else statements could be pulled out to different functions, so you do
```
if type == PaymentTransaction:
    get_payment_transaction_items(self)
elif type == PaymentRelaton:
    get_payment_relation_items(self)
```
instead of this huge mess :zipper_mouth_face: :smile: 

Another idea is to make each payment type of object expose `get_title`, `get_timestamp`, `get_description`, … so you can get each of the properties by doing `title = self.get_title()` instead of the whole if/else thingymajig.

PR #2019, opened 2018-02-04
Social event attendance list

KevinMidboe:
Add feature attendance list for social events.

- [ ] QA / Code Review

## Code Checklist

## Description of changes
Added attendance list for social events. Now when viewing a social event a button will be displayed for showing attending users.  
A user can choose a default setting in profile -> personvern where it will set the default state for showing on these lists. The default value is False. A user can also choose whether to be added or retracted from the list of a specific event.


templates/events/details.html
```diff
@@ -393,6 +418,55 @@ <h3>Meld deg av {{ event.title }} </h3>
...
+                        </div>
+                        <div class="modal-footer">
+                            {% if user_attending %}
+                                {% if show_as_attending %}
+                                    <a href="{% url 'toggleShowAsAttending' event.id %}"><button class="btn btn-warning">Ikke vis meg på denne listen</button></a>
+                                {% else %}
+                                    <a href="{% url 'toggleShowAsAttending' event.id %}"><button class="btn btn-success">Gjør meg synlig som påmeldt</button></a>
+                                {% endif %}
+                            {% else %}
+                                <button class="btn btn-success disabled">Meld deg på arrangementet</button>
```
KevinMidboe: Siden users som ikke er meldt på arrangementet ikke kan velge om de skal være synlige eller ikke, la jeg til en disabled knapp så det ikke bare var en lukk knapp for modulen. Er det nødvendig å ha en disabled knapp for users som ikke er meldt på arrangementet enda?

apps/profiles/migrations/0003_privacy_visible_as_attending_events.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.8 on 2018-02-04 20:41
```
s0lvang: Delete this

apps/events/migrations/0017_auto_20180204_2141.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
KevinMidboe: Okokokokokok

apps/events/models.py
```diff
@@ -847,6 +860,13 @@ def delete(self):
     def is_on_waitlist(self):
         return self in self.event.waitlist_qs
 
+    def is_visible_as_attending(self):
+        # Is visible as attending if default users privacy settings are set and
+        # explicit settings for attending this event does not override privacy setting
+        user_setting = self.user.get_visible_as_attending_events()
+
+        return (user_setting and self.show_as_attending_event is None) or self.show_as_attending_event
```
sklirg: `show_as_attending` sounds like a variable which should be `True` or `False`, not `None`.

templates/events/details.html
```diff
@@ -393,6 +418,54 @@ <h3>Meld deg av {{ event.title }} </h3>
...
+                                    {% for user in event.attendance_event.visible_attendees_qs %}
+                                    <tr>
+                                        <td>{{ user.first_name }}</td>
+                                        <td>{{ user.last_name }}</td>
+                                    </tr>
+                                    {% endfor %}
+                                </tbody>
+                            </table>
+                             {% if user_setting_show_as_attending %}
+                            <span><i>Du vises automatisk som påmeldt for alle sosiale arrangement, dette kan endres under personvern på <a href='/profile/privacy'>profilen din her.</a></i></span>
```
sklirg: Use the Django URL template tags for URLs, like `{% url 'profile_index' %}` or whatever the name is (check urls.py for profile).

templates/events/details.html
```diff
@@ -393,6 +418,54 @@ <h3>Meld deg av {{ event.title }} </h3>
...
+                            {% else %}
+                            <span><i>Hvis du ønsker å automatisk vises som påmeldt på alle fremtidige sosiale arrangement kan det gjøres under personvern på <a href='/profile/privacy'>profilen din her.</a></i></span>
```
sklirg: Use the Django URL template tags for URLs, like `{% url 'profile_index' %}` or whatever the name is (check urls.py for profile).

templates/events/details.html
sklirg: Could you post a screenshot with what you mean?

I don't think the "Meld meg på"-button should be visible at all if you can not attend.

KevinMidboe:
Screenshots of event page with the new 'vis påmeldte' button.

## Extras
extras_new

## Payment
payment_new

## Payment w/ extras
extra payment_new

apps/events/models.py
KevinMidboe: `None` is used to check if the user has explicitly set a value for this specific event. 

If the value of `show_as_attending_event` is `None` we know that a user has not yet set a preference for this event, and the default value set in the users privacy setting with the button *Synlig på alle sosiale arrangement* is used to find out if the user should be displayed in the list or not. If the value is `False` or `True` we know that the user has specifically set a value for this event.

This way the default value set in the users privacy settings is used by default and if the user wishes be shown/hidden from this specific event the `show_as_attending_event`is changed from `None` to `True` or `False` respectively.

templates/events/details.html
KevinMidboe: I agree. I think the disabled *Meld deg på arrangementet* button creates confusion and should be removed.

disabled_button

apps/events/models.py
sklirg: I'd think that setting the value to the users specified default _when the user attends_ would achieve the same result.

Attendee.attend() pseudocode:
`attendee.show_as_attending_event = attendee.user.privacy.show_as_attending_events`

This would set `attendee.show_as_attending_event` to the specified default value in the user's privacy settings when the user attends.

This would also make it so that changing the privacy setting does not change the lists for all past events for which you have _not_ opted in to show your attendance.

apps/profiles/migrations/0003_privacy_visible_as_attending_events.py
sklirg: Why delete this? It's required to migrate the database from one state to another.

apps/events/migrations/0017_auto_20180204_2141.py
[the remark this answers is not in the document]
sklirg: Why delete this? It's required to migrate the database from one state to another.

Not sure why there are more fields than the `show_as_attending`-one, though. Those should not be necessary.

PR #2025, opened 2018-02-25
Delete old genfors app

duvholt:
Didn't even use nodejs as backend. Literally old.


sklirg:
This does not remove database tables. Is that intentional? Do we want lingering database tables or should we drop them? If so, should we manually drop them in psql or tell Django to drop them for us (we could probably "unmigrate" (roll back) migrations back to initial, unless something is dependant on them)?

duvholt:
Doesn't really matter to me. It could be fun to have the old data for historical reasons. 
I don't see any reason to create Django migrations for this though.

Issue #2028, opened 2018-03-05
"Show attendees"-button does not show on events you cannot attend


MrWe:
## What kind of an issue is this?

- [x] Bug report
- [ ] Feature request

## What is the expected behaviour?
"Show Attendees"-button is shown on all social events. 

## What is the current behaviour?
"Show Attendees"-button is shown on only social events where the user can sign up for the event. 
The attendees list is still in the HTML, but not shown on the page. 

## How do you reproduce this problem? 

Enter an event where you cannot sign up, for example an event for 5. graders from a user in the 4. grade. 

## Other information

<img width="459" alt="screenshot 2018-03-05 22 18 19" src="https://user-images.githubusercontent.com/6748603/37000323-5bc7b810-20c3-11e8-88fe-e4d489538be3.png">
<img width="459" alt="screenshot 2018-03-05 22 18 27" src="https://user-images.githubusercontent.com/6748603/37000324-5beb3da8-20c3-11e8-946d-d1b2e21e780d.png">

sklirg:
Hi,

Thanks for this issue. This is quite critical, and in the future we'd prefer if information about issues like these were sent privately by email.

sklirg:
Fixed in #2031 and #2033.

PR #2033, opened 2018-03-07
Revert "Simply remove modal for attendees attending this event"

KevinMidboe:
Reverts dotkom/onlineweb4#2031

### DIFF
--- a/templates/events/details.html
+++ b/templates/events/details.html
@@ -181,6 +181,10 @@
                                                             <div class="action col-md-6 col-sm-6 col-xs-6">
                                                                 <a href="#refund" role="button" class="btn btn-large btn-danger" data-toggle="modal">Refunder og meld meg av</a>
                                                             </div>
+
+                                                            {% if event.event_type is 1 %}
+                                                            <div class="action col-md-6 col-sm-6 col-xs-6"><a href="#attendees" role="button" class="btn btn-large btn-info" data-toggle="modal">Vis påmeldte</a></div>
+                                                            {% endif %}
                                                         </div>
                                                     {% else %} <!-- User has paid flag set but no payment_relation object -->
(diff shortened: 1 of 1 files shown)


KevinMidboe:
Re-adding the modal for showing other attendees for **only** social events. 
This should close the issue #2028 as it addresses the issue where even if the button for showing attendees was not shown, the list of attendees was still rendered in the HTML code. 

New behaviour: Check if event type is social event and the user is either attending or is able to attend the event. 
The modal render condition: `if event_type is social and is_able_to_attend or is_attending`

templates/events/details.html
```diff
@@ -403,6 +418,54 @@ <h3>Meld deg av {{ event.title }} </h3>
                 </div>
             </div>
 
+            {% if event.event_type is 1 and user_status.status or user_attending %}
```
duvholt: Don't use `is` for comparing numbers in Python.
`is` tests if the objects compared are the same. It does work for small integers(up to 256), but this is bad practice as it only works because small integers are mapped to the same object in CPython.

templates/events/details.html
```diff
@@ -403,6 +418,54 @@ <h3>Meld deg av {{ event.title }} </h3>
...
+            <div id="attendees" class="modal fade">
+                <div class="modal-dialog">
+                    <div class="modal-content">
+                        <div class="modal-header">
+                            <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+                            <h3>Påmeldte til {{ event.title }} </h3>
+                            <h5><i>Siden dette er et sosialt arrangement kan du ønske å vise deg som påmeldt for andre å se.</i></h5>
```
duvholt: h5 shouldn't be (ab)used to change font size here. 
I'm not sure how this looks rendered, but I would create a new class (model-subtitle-ow? idk) instead.

PR #2051, opened 2018-04-12
Fix/uniqueify rfid

### LINKED ISSUE #1919: Attendance API tries to query attendees by RFID value "None"

sklirg:

- QA / Code Review

## Code Checklist

    - Weell, kinda not really, if someone abused the fact that RFIDs were not unique. Other than that though, we're fine.

## Description of changes

Made RFID field on User model unique, so that users cannot have the same RFID as someone else. We actually expect the field to be unique, so the constraint added here makes sure it actually is unique.

Should probably fix #1919.


sklirg:
@Aborysa this fix adds a new `status_code`. Could you confirm that this does not break regme in any way? :-D

This fix also catches another exception which should make the app return JSON responses for all events.

sardred:
Do you mean the new [`attend_status`](https://github.com/dotkom/onlineweb4/blob/019673148b386efcbaafe8dfb6db4d7520f2c58b/apps/events/views.py#L457)? Would be nice to add it into the overview [here](https://github.com/dotkom/regme/blob/master/app/src/services/attendee/attendee.service.js#L10), but from what I can gather the only `attend_status`-specific code I can find is [this](https://github.com/dotkom/regme/blob/master/app/src/components/registration/index.js#L100). If you want your new response to fall under that same case, add code 51 to the if-statement.
I suppose if you want to follow the scheme we have, the code should be 60 instead of 51.

sklirg:
I meant `attend_status`, yes. 😶 

Looks like the messages just passes through, so the status code isn't very relevant.

I thought the `attend_status` codes were 4x for user errors and 5x for "server errors", but oh well.

PR #2070, opened 2018-05-14
Fix inconsistent permission checking in events

torjusti:

- QA / Code Review

## Code Checklist

## Description of changes

This adds a previously breaking test and fixes the issue behind it. If a committee created an event of a type not belonging to that committee, such as bedkom creating a social event, members of that comittee would not be able to export lists for that event, or send mails about the event.

### DIFF
--- a/apps/events/tests/view_tests.py
+++ b/apps/events/tests/view_tests.py
@@ -14,14 +14,15 @@ from apps.authentication.models import AllowedUsername
 from apps.payment.models import PaymentDelay
 
 from ..models import TYPE_CHOICES, AttendanceEvent, Event, Extras, GroupRestriction
-from .utils import (add_payment_delay, add_to_arrkom, add_to_trikom, attend_user_to_event,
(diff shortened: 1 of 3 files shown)


apps/events/tests/view_tests.py
```diff
@@ -559,6 +560,17 @@ def test_missing_access(self):
...
+    def test_get_own_social_event_as_bedkom(self):
+        add_to_bedkom(self.user)
+        bedkom = Group.objects.get(name__iexact='bedkom')
+        event = generate_event(TYPE_CHOICES[0][0], organizer=bedkom)
+        url = reverse('event_mail_participants', args=(event.id,))
+
+        response = self.client.get(url)
+
+        self.assertEqual(response.context['event'], event)
+        self.assertEqual(len(mail.outbox), 0)
```
sklirg: Not sure why this test function should test that no emails are sent. I can see that some others do, but it should not be necessary.

PR #2071, opened 2018-05-16
Add an IsLoading-component for lengthy cross-domain requests

### LINKED ISSUE #1794: Make user stay at "Refill Saldo" page after paying
Currently, after the Stripe popup closes, there is a wait of a couple 100 ms before a POST request is initiated. We should open a modal or some kind of loading message saying "Processing payment... please wait" while we POST the response from Stripe, to process it in the backend.

sklirg:

- QA / Code Review

## Code Checklist

    - CI does not, though. Seems unrelated since I haven't changed any backend code.

## Description of changes

Adds a component which tells the user that the server is processing their request. This can be used when we require (or expect) the user to stay on the page after having done an action, if we do a request and want to report back to the user. Examples of such actions are payments and g suite-related actions, which contact a third-party service.

The component registers some eventListeners which makes it easy to dispatch the actions from vanilla JS as well.

This PR implements the actions for (un)subscribing to the mailing lists as well as the "saldo" function on my profile.

Fixes #1794.

image

Showcasing the popup message.

---

image

Message location in a full HD browser

---

image

Message on a phone

### DIFF
--- a/assets/core/LoadingComponent/IsLoading.js
+++ b/assets/core/LoadingComponent/IsLoading.js
@@ -1,0 +1,91 @@
+import React from 'react';
+import Transition from 'react-transition-group/Transition';
+
+import style from './style.less'; // eslint-disable-line no-unused-vars
+
+const duration = 400;
+
+const defaultStyle = {
+  transition: `opacity ${duration}ms ease-in-out`,
+  opacity: 0,
+};
+
+const transitionStyles = {
+  entering: { opacity: 0 },
+  entered: { opacity: 1 },
+};
+
+class IsLoading extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      show: false,
+    };
+    this.registerListeners();
+  }
+
+  registerBeforeUnloadListener() {
+    window.addEventListener(
+      'beforeunload',
+      (event) => {
+        if (this.state.show) {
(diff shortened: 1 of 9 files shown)


sklirg:
I think we should consider if this should be even more intrusive. It's nice to inform the user about the requests as we do now, but because we sometimes require the users to stay, it might be interesting to look into adding a modal of some sort which "disables" the web page. This might be a bit bad for user experience, but the server actually requires it as of now.

Another thing to consider is refactoring all of these processes; deferring the requests and respond to them asynchronously. This can be done by using message queues, where we simply say "yup we'll fix this for you don't you worry" and then instantly respond to the request after adding the task to a queue. The task can happen at any time, either instantly or in 5 minutes time. This does require some larger refactoring, but it can be implemented iteratively, starting small for testing purposes and refactor all events to use it later. What's important to keep in mind if this is to be implemented is to handle any failing tasks and inform the user, by email or whatnot. Such as failed to process a payment or failing to subscribe a user to a list. Anywho, this is quite off-topic for this PR :^)

duvholt:
How about using [onbeforeunload](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload) to prevent the user from navigating away by accident?

sklirg:
Great suggestion, I added it.

PR #2074, opened 2018-05-17
Proof of concept for running async tasks using celery

duvholt:

- QA / Code Review
- Proof of concept

## Code Checklist

- [x] These changes requires changes to configuration in production :skull_and_crossbones: :warning:

## Description of changes
Proof of concept of running async tasks. 
Added a celery task for updating gsuite mailing lists.

### DIFF
--- a/apps/gsuite/mail_syncer/signals.py
+++ b/apps/gsuite/mail_syncer/signals.py
@@ -1,15 +1,10 @@
-import json
 import logging
 
-from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.db.models.signals import pre_save
 from django.dispatch import receiver
-from googleapiclient.errors import HttpError
 
-from apps.gsuite.mail_syncer.utils import insert_email_into_g_suite_group
-
-from .main import remove_g_suite_user_from_group
+from .tasks import update_mailing_list
 
 User = get_user_model()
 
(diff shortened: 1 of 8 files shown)


sklirg:
Nice work.

Have you tested how exception handling works with regards to default Django settings? Will emails be sent to admins if a task raises an exception during execution? Is the logging caught by Django logging configuration?

I've used Celery before and it works nice. We used it with RabbitMQ instead of Redis as the queue, have you done any research with regards to what and how and why Redis?

Have you thought anything about how the broker should be hosted? Similar to how mommy is hosted currently, maybe?

duvholt:
>Have you tested how exception handling works with regards to default Django settings? Will emails be sent to admins if a task raises an exception during execution? Is the logging caught by Django logging configuration?

For development exceptions will propagate as normal. Not really sure how logging is handled when running in Celery, but it is written to stdout at least.

>I've used Celery before and it works nice. We used it with RabbitMQ instead of Redis as the queue, have you done any research with regards to what and how and why Redis?

I don't think it matters which broker we use. I chose Redis because it is written in C while RabbitMQ is written in Erlang and if I remember correctly has a whole bunch of dependencies. 

>Have you thought anything about how the broker should be hosted? Similar to how mommy is hosted currently, maybe?

Similar to mommy yeah.

oleast:
I've had a few looks at this at different times, and I end up with thinking I don't know quite enough to give an all clear on this one.

What would be the changes made to production to make it work, is it just setting up Redis?
Should Redis also be set up on the docker configuration for development?

I do see that Celery could be quite useful, so it would be nice to have this merged :)

duvholt:
It's been a while since I worked on this, but from what I remember the production changes required are:

- Setting `OW4_CELERY_TASK_ALWAYS_EAGER` to false
- Setup redis
- Run the celery as a separate services like mommy

Since the default behavior is the run tasks in blocking mode it's not necessary to do any changes to the development environment.

PR #2076, opened 2018-05-28
Enable documentation for postgraduate studies

martinlunde:
Feature/fix

## Code Checklist

## Description of changes
Add description and documentation field to forms when applying for postgraduate membership.

### DIFF
--- a/assets/profiles/dependent_fields.js
+++ b/assets/profiles/dependent_fields.js
@@ -1,12 +1,23 @@
 import $ from 'jquery';
 
 export default () => {
-  $('#field-of-study-application-documentation').hide();
+  let master_doc = $('#field-of-study-application-master-documentation');
+  let social_doc = $('#field-of-study-application-documentation');
+  social_doc.hide();
+  master_doc.hide();
+
   $('#id_field_of_study').on('change', () => {
-    if ($('#id_field_of_study').val() === '40') {
-      $('#field-of-study-application-documentation').show();
-    } else {
-      $('#field-of-study-application-documentation').hide();
+      let field_of_study = parseInt($('#id_field_of_study').val());
+
+    if (field_of_study >= 10 && field_of_study <= 30){
+      social_doc.hide();
+      master_doc.show();
+    } else if (field_of_study === 40){
+      social_doc.show();
+      master_doc.hide();
+    } else{
+      social_doc.hide();
+      master_doc.hide();
     }
   });
 };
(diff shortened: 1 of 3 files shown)


templates/profiles/membership.html
```diff
@@ -57,6 +57,12 @@ <h3>Medlemskap</h3>
                             studentweb.</p>
                         {{ field_of_study_application.documentation|as_crispy_field }}
                     </div>
+                    <div id="field-of-study-application-master-documentation">
+                        <p>For å søke master medlemsskap i linjeforeningen, kreves det at det gis et bevis på opptak.
```
sklirg: "Medlemskap" is the correct way to write this in Norwegian[0] :D

Also, "master medlemskap" is a nice "orddelingsfeil", so maybe even rewrite it to "medlemskap som masterstudent"?

You can probably also drop the comma in this sentence to make it flow a bit better.

0: http://www.sprakradet.no/svardatabase/sporsmal-og-svar/medlemskap-eller-medlemsskap/

templates/profiles/membership.html
```diff
@@ -57,6 +57,12 @@ <h3>Medlemskap</h3>
...
+                            Dette må dokumenteres ved at du laster opp f.eks et screenshot av din studieplan på studentweb.
```
sklirg: If we're gonna be real Norwegians here we could name it "skjermdump", but that's up to you.

Anywho, "f.eks." needs a dot after "eks." too.

We could also be helpful and link to studentweb.ntnu.no.

While we're on it, we should probably coordinate with Hovedstyret regarding what's accepted as proof in case more options are available.

assets/profiles/dependent_fields.js
```diff
@@ -1,12 +1,23 @@
...
+
+    if (field_of_study >= 10 && field_of_study <= 30){
+      social_doc.hide();
+      master_doc.show();
+    } else if (field_of_study === 40){
+      social_doc.show();
+      master_doc.hide();
+    } else{
+      social_doc.hide();
+      master_doc.hide();
```
sklirg: Personal preference probably, but I'd rather hide it by default and show it conditionally than hide and or show it conditionally.

```javascript
social_doc.hide()
master_doc.hide()
if master study
    master_doc.show()
else if social study
    social_doc.show()
```

And maybe even hide the HTML element by default (using CSS) instead of doing two DOM operations which hide or show the element (jQ might be efficient enough to only be done in one operation, but I'm not sure).

```javascript
if master study
    master_doc.show()
else if social study
    social_doc.show()
```

Both ways are cleaner code nonetheless.

martinlunde:
This PR should now be redundant with the introduction of authentication through feide.

PR #2078, opened 2018-07-03
Add image field to Hobby and Resource API endpoints

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

This pull request adds the image field to the `resourcecenter` and `hobbygroups` API endpoint.
This is done to make the API compatible with 'Onlineweb Frontend'.

Also added a newline to appease stylelint.

## Update:

Changed image field on resource and hobbys to use `ResponsiveImage`.
Added an image type to use for resources and hobbys.


oleast:
From what I can see, that would require the models to be updated to have the foreign key of a 'ResponsiveImage', rather than the 'ImageField' they use now.

It can be done, but will require some changes in prod.

sklirg:
Interesting. I'd vouch for that being done, so all images are handled the
same.

MrHanSan:
@oleast could you do a rebase, or merge inn develop? Will have another look.

oleast:
It has been rebased now, but I am a bit unsure of how we will have to do this in production?
Will we have to add a manual migration for setting a default ResponseiveImage or something like that?

sklirg:
Correct. Everything will start with no image, so you'll have to upload a new image to each object.

PR #2081, opened 2018-07-22
Feat/make migration merge

Aborysa:

- QA / Code Review

## Code Checklist

## Description of changes
Added new make command for merging migrations and for installing node packages.


sklirg:
We already have yarn.lock, so we shouldn't add npm's lockfile.

Also, what does this migration change and why is it required?

PR #2084, opened 2018-07-29
add fallback to local jquery if CDN fails

### LINKED ISSUE #2054: Consider serving JS dependencies locally

- Enhancement (suggestion)

## Other information

The jQuery CDN was recently misconfigured for SSL/TLS (https://github.com/jquery/codeorigin.jquery.com/issues/34), which resulted in pretty much nothing working on our site due to it expecting a global jQ object to work with.

Even though we are moving away from jQ, I suggest serving the jQ lib locally so we can avoid issues like this in the future. This should be quite a quick fix to solve a potentially massive problem, if it ever happens again.

To reproduce/simulate the problem, load the webpage and block the request from `code.jquery.com`. 
This can be done in Chrome by going to Developer Tools -> Network -> right click the script loading from code.jquery.com and click "Block request URL".

aslakhol:

Resolves #2054 by adding a fallback to local jquery. 

## Code Checklist


sklirg:
For people who've heard the horror about `document.write()`:

> If the document.write() call is embedded within an inlined HTML <script> tag, then it will not call document.open(). 

- https://developer.mozilla.org/en-US/docs/Web/API/Document/write (thanks @duvholt)

PR #2092, opened 2018-08-15
Fix/dataporten registration update user

aslakhol:
Fixes an issue with the dataporten integration. New users would not have their memberships or profiles altered when their applications where approved.

This also adds ntnu_username to account automatically when applying for membership

## Code Checklist


aslakhol:
I will likely go over this to see if it needs refactoring tomorrow morning, but it is late.

Issue #2075, opened 2018-05-18
Articles dashboard page redirects to login instead of sending 403


torjusti:
## What kind of an issue is this?

- Bug report

## What is the expected behaviour?

If you attempt to visit the articles-path in the dashboard without the required permissiosn, you should receive a 403 error.

## What is the current behaviour?

You're redirected to the login page.

## How do you reproduce this problem? 

Attempt to visit https://online.ntnu.no/dashboard/article/ without the required permissions. You will be redirected to the login page. You can also find the link on the dashboard home page, in the recently activated modules section - this probably shouldn't happen, either :)

## Other information

This is probably due to decorators being used in the article views without setting flags correctly.

sklirg:
Looks like a really simple fix. Add `raise_exception` to the permission required decorator, as detailed here: https://docs.djangoproject.com/en/2.0/topics/auth/default/#the-permission-required-decorator

PR #2098, opened 2018-09-10
Fixed issue #2075

tokongs:
Added a return 403 to the permission_required decorator in the view for
dashboard/article.

- QA / Code Review

## Code Checklist

## Description of changes
Made some changes to the permission_required decorator in /article/dashboard/view.py so that it return a 403 when the user doesn't have the required permissions. 

## P.S
First ever pull request... pls don't kill me...

### DIFF
--- a/apps/article/dashboard/views.py
+++ b/apps/article/dashboard/views.py
@@ -14,7 +14,7 @@ from apps.article.models import Article
 from apps.dashboard.tools import check_access_or_403, get_base_context
 
 
-@permission_required('article.view_article')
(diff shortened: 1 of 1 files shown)


MrHanSan:
We don't really need the ```check_access_or_403`` check, but it works a bit differently than the decorator. The ```check_access_or_403``` method checks if the user is in the group "Komiteer" or if the user is a super user, while the decorator only checks for permissions, regardless of groups.

Removing ```check_access_or_403``` should be OK, but might cause some edge cases(?)

PR #2107, opened 2018-09-13
fixed attendeebug

s0lvang:
This pull-request fixes the bug where you could see attendees at company-events.

I just added an extra if-sentence, but because of intendation it seems like more.


KevinMidboe:
🥇 Looks good to me too.

PR #2109, opened 2018-09-14
Added settings to not escape mailto links in markdown from chunks

### LINKED ISSUE #1855: Mailto links don't work in articles and chunks

- [x] Bug report
- [ ] Feature request

Mailto links should work

Mailto links link to /# instead of opening email

Just look at any of the mailto links in the chucks, or in an article

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

Fixes #1855 

[default settings](https://github.com/trentm/django-markdown-deux/blob/master/lib/markdown_deux/conf/settings.py) for `django-markdown-deux` is to escape mailto links.

This change lets markdown render `unsafe` elements.

### DIFF
--- a/onlineweb4/settings/__init__.py
+++ b/onlineweb4/settings/__init__.py
@@ -10,6 +10,7 @@ from onlineweb4.settings.logging import *
 from onlineweb4.settings.raven import *
 from onlineweb4.settings.rest_framework import *
 from onlineweb4.settings.stripe import *
+from onlineweb4.settings.markdown import *
 
 try:
     from onlineweb4.settings.local import *
--- a/onlineweb4/settings/markdown.py
+++ b/onlineweb4/settings/markdown.py
@@ -1,0 +1,8 @@
+MARKDOWN_DEUX_STYLES = {
+    'default': {
+        "extras": {
+            "code-friendly": None,
+        },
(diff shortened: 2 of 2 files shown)


sklirg:
Is there a way to whitelist elements rather than turning off safe mode globally for all markdown? I'm not that familiar with what you technically can achieve with markdown in terms of XSS etc., so if this is the only possible solution we should be aware of the risks.

oleast:
It may or may not be possible to do with a [link pattern](https://github.com/trentm/python-markdown2/wiki/link-patterns)
Not sure if i understand correctly about the [default behavior](https://github.com/trentm/python-markdown2/wiki/link-patterns#default-link-behaviour), if only `http(s)` and `ftp` link are supported, or that you could write a link pattern which makes it possible.

sklirg:
Looks like there's an open issue regarding this: https://github.com/trentm/python-markdown2/issues/257

So yeah, extending the link pattern to include mailto links sounds like the best choice in my opinion.

sardred:
ahahahh yeah man me too i love link patterns

@oleast, do we "extend the link pattern to include mailto links", accept the potential risks of merging this, or close it?

sardred:
Better safe than sorry! Closing and deleting branch.

PR #2124, opened 2018-09-27
Update dashboard default placeholder for datepicker, datetimepicker a…

emilps:
…nd timepicker.

- QA / Code Review

## Code Checklist

- [X] The code follows dotkom code style 
- [X] The code passes the defined tests
- [X] The code is ready to be merged

## Description of changes
Changed default placeholders because they didn't fit all purposes. 

Example: Create feedback on event in dashboard.  

bedriftspresentasjon_med_oath_-_dashboard_-_online

As seen on the picture the text fields are contradictory.

### DIFF
--- a/apps/dashboard/widgets.py
+++ b/apps/dashboard/widgets.py
@@ -108,7 +108,7 @@ class DatePickerInput(TextInput):
             final_attrs['value'] = ''
 
         # Kept for backwards compatibility with existing forms.
-        final_attrs['placeholder'] = 'Den skal vises fra ...'
+        final_attrs['placeholder'] = 'Vennligst velg en dato ...'
         if attrs.get('placeholder', False):
             # Update the placeholder text if supplied.
             final_attrs['placeholder'] = force_text(attrs.get('placeholder'))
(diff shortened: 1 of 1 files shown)


sklirg:
This changes the default placeholder value of the widget. It should probably be changed where it's used, i.e. somewhere around here: https://github.com/dotkom/onlineweb4/blob/develop/apps/events/dashboard/forms.py#L103

sklirg:
Oh right! My bad. Because of the image, I thought you wanted to change it for only the "add feedback"  form. I now see that you want to update the _default value_.

This change would be nice, however, as noted in the preceding comment, it's kept in for backwards compatibility with some form fields. It would be nice if these "old" form fields could be updated with the new syntax. I think it's currently in use in the posters app, so somewhere around here: https://github.com/dotkom/onlineweb4/blob/develop/apps/posters/forms.py#L46

image

sklirg:
On third thought though, maybe the default text makes sense since the label for the field in question describes what it should be filled with, so the placeholder can tell you what to input.

Okay, I agree with the change as it currently stands.

Issue #2126, opened 2018-10-02
ITEX should be able to see feedbacks from excursions


oleast:
## What kind of an issue is this?

- Bug report

## What is the expected behaviour?

Members of the ITEX group shoudl be able to see the feedbacks from their events.

## What is the current behaviour?

Members of the ITEX group can not see feedbacks from excursions, because it is hard-coded that only 'eksKom' can do that.

## Other information

The offending lines of code can be found [here](https://github.com/dotkom/onlineweb4/blob/0e7e3e2fd59ff5a3b77e9404b166fcb7716b0212/apps/events/utils.py#L31-L53)

oleast:
Resolved by #2131

Issue #1209, opened 2015-10-02
Revamp how we handle committees


sklirg:
We have multiple apps that needs to know what committee(s) a user is part of, so we need some way to figure out what (and/or which) committee(s) a user is in.

Some suggestions:
- permission class on committee groups, then find which groups have this permission (_hacky_)
- extend django.contrib Groups

hernil:
#1308

hernil:
I'd like anyone interested in modelling this to poke me so we can get a group working on this next week.

myth:
Recap of the group discussion:

Since django does not support swappable group object out of the box, we should attempt to subclass Group and use it where we need, for example by using a custom manager.

An even better approach, at least in terms of functionality, is to set up our base group (see list of bullets below) by injecting common fields into the already existing Group object. Then django will add them to the same database table, removing one join operation from the equation. And we do really need to keep those in mind. A possible example of how this might be set up in models:

```
if not hasattr(Group, 'builtin'):
    setattr(Group, 'builtin', models.BooleanField('Gruppen er en innebygd gruppe eller tilknyttet kjernefunksjonalitet.')
```

The clear advantage of this compared to subclassing is the need to add another JOIN layer for our organizational entities, as well as laying the foundation for other future usages that depend on similar functionality.

Some bulletpoints from the discussion:

* Group description, short and long.
* Description should include the organizational entity's purpose and intent (from the reworked committee guidelines)
* A tree-like structure of organizational nodes, allowing inference of responsibility, ownership, informational aggregation, automatic organizational charts, automatically generated E-mail lists etc. Limitless possibilities.
* Remove all hacks with ints and strings in our code / settings
* A notion of parent and child branches (ref tree-like structure)
* Application for a group can be a generic group application form built upon the approvals app maybe? Automatic committee membership applications, prioritizing, application handling and statistics anyone?
* Fields for leader, second in command and possibly responsible for economy? An obvious solution would be an M2M field for unlimited roles/positions. We should keep the amount of M2M fields in consideration. 6-8 levels of joins are expensive. Hopefully it shouldn't be a big problem. Some benchmarking with 50-100 group objects with the proposed design would be nice.
* User-group-roles could allow automatic generation of "positions" on the user profile.
* User-group-roles could allow automatic reminders of who is up for "long and faithful service" medals, 3-year committee reapproval limit etc.
* And a plethora of other possibilities

sklirg:
[...]

PR #2131, opened 2018-10-03
Allows ITEX to view feedback from excursions.

R33an1:

Bug fix 

- QA / Code Review

## Code Checklist

## Description of changes
As stated in title.  #2126

### DIFF
--- a/apps/events/utils.py
+++ b/apps/events/utils.py
@@ -51,7 +51,7 @@ def get_types_allowed(user):
             types_allowed.append(5)  # ekskursjon
 
         if group.name == 'itex':
-            types_allowed.append(5) # ekskursjon
+            types_allowed.append(5)  # ekskursjon
 
     return types_allowed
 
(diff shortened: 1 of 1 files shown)


oleast:
As it stands right now, Ekskom has access to all that for ITEX already?
It was mostly done this way to keep it with how it is done previously.

As this was resolved by giving the results directly to the responsible person in ITEX as a PDF, we can revert this, and hope that we get to fixing permissions before ITEX needs access next year?

sklirg:
Good catch, that's correct. I think it has been circumvented before by not
using "Ekskursjon" for ITEX-events, though.

Also, currently all events and their permission (for the rest of the site)
goes through permissions rather than group membership, so that might be a
more interesting approach if this is to be refactored.

oleast:
I think we would like to do that as part of the #1209.
Switching everything that can be switched to permissions, to keep things more consistent and manageable.
I will set up a pull request to revert this one.

Issue #2128, opened 2018-10-03
We need better logging of people being removed from events. 


aslakhol:
## What kind of an issue is this?

- Bug report

## What is the expected behaviour?
We need to know who removed a person, which person and what time. 

## What is the current behaviour?
We don't quite know what we have logged.

sklirg:
I'd suggest also implementing an email notification to the organizing committee if a person is unattended after the `unattend_deadline`.

oleast:
Fixed by #2138

aslakhol:
Does #2138 log when users are removed from events through the adminpanel? @martinlunde

martinlunde:
Unfortunately no, however the option of deleting attendees is removed from the adminpanel, so it should not cause any problems.

PR #2137, opened 2018-10-09
Log user removing other users from events #Hacktoberfest #2128

martinlunde:
Feature request in order to provide more accurate logging of unwanted events.

## Code Checklist

## Description of changes
You should now be able to view logs of users removing attendees of an event from the dashboard.


martinlunde:
#2128

sklirg:
Note that this will only work when people are removed through the dashboard. I'd suggest implementing something similar for the administration panel, maybe even a generic `Attendee.unattend()` function which can take an administrator as a parameter.

Issue #1956, opened 2017-11-22
Better feedback in dashboard when creating a careeropportunity


Westermoen:
## What kind of an issue is this?

- [x] Bug report
- [x] Feature request

## What is the expected behaviour?
Have better feedback in the dashboard when failing to create a careeropportunity

## What is the current behaviour?
Little or no feedback when failing to create a careeropportunity

## How do you reproduce this problem? 
Insert wrong data or break the rules defined for one of the input fields

oleast:
Fixed by #2144

PR #2144, opened 2018-10-13
Set a max length for two fields in the form for adding career opportunities

tokongs:

- QA / Code Review

## Code Checklist

- [x ] The code follows dotkom code style 
- [x ] The code passes the defined tests
- [ x] The code is ready to be merged

- [ x] The changes are backwards compatible

## Description of changes
Issue #1956 calls for better feedback when creating a career opportunity in the dashboard, so I added a max length to the title field and the ingress field as I think that will mitigate the need for better feedback. I don't really see what other wrong data you could enter as the form is now.

### DIFF
--- a/apps/careeropportunity/forms.py
+++ b/apps/careeropportunity/forms.py
@@ -5,9 +5,9 @@ from apps.careeropportunity.models import CareerOpportunity
 
 class AddCareerOpportunityForm(forms.ModelForm):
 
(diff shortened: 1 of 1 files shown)


sklirg:
I think Django wants a migration for any changes to a model field, would you mind double-checking if `makemigrations careeropportunity` creates one, and if so, add it?

PR #2148, opened 2018-10-17
Page size parameter for the rest api

Aborysa:
- QA

## Code Checklist

## Description of changes
Added a custom paginator for the rest api framework that allows a request to set a custom page size using the `page_size` parameter. This allows use to reduce the amount of api calls.
Also set the default paginator to the new custom one, but kept default page size to 10 in order to not break apps which might assume the default is always 10.

### DIFF
--- a/utils/pagination.py
+++ b/utils/pagination.py
@@ -4,7 +4,7 @@ from rest_framework.pagination import PageNumberPagination as RFPageNumberPagina
 class PageNumberPagination(RFPageNumberPagination):
     page_size = 10
     max_page_size = 80
-    
+
     def get_page_size(self, request):
         if ('page_size' in request.query_params) and request.query_params['page_size'].isnumeric():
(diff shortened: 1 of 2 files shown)


utils/pagination.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework.pagination import PageNumberPagination as RFPageNumberPagination
+
+
+class PageNumberPagination(RFPageNumberPagination):
+    page_size = 10
+    max_page_size = 80
+    
+    def get_page_size(self, request):
+        if 'page_size' in request.query_params:
+            return min(self.max_page_size, int(request.query_params['page_size']))
```
sklirg: What if someone tries to supply `page_size` as a string? => `ValueError`

Use `"example string".isnumeric()` to check if it is a number before you try casting it 👍

PR #2150, opened 2018-10-17
[WIP]Search in chunks api

Aborysa:
- QA

## Code Checklist

## Description of changes

Added prefix- search option to chunks api so it is possible to create chunks for specific apps by using a prefix, i.e `regme_`, `nibble_`, or `opptak_` and then only fetch app specific chunks by added a search parameter `?search=app_`

### DIFF
--- a/apps/chunksapi/views.py
+++ b/apps/chunksapi/views.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
 
 from chunks.models import Chunk
-from rest_framework import mixins, viewsets, filters
-from rest_framework.views import APIView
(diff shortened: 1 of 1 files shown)


sklirg:
I'm think we're trying to drop chunks because it's not maintained any more and has had some bugs with recent Django versions. I'd suggest creating something new which does the same thing (or fork and fix 4realz), primarily focused on an API rather than Django templates, if the functionality is still wanted.

Aborysa:
Sounds like a good idea 👍

torjusti:
I'm closing this as we want to move away from chunks and this PR probably won't be relevant then. Feel free to re-open the pull request if I'm wrong.

Issue #2052, opened 2018-04-12
Attend API responds with "card # does not exist" if the user isn't actually attending


sklirg:
## What kind of an issue is this?

- Bug report

## What is the expected behaviour?

If a user who isn't registered as attending an event tries to attend an event the system should state that the user isn't attending the event. (_wow such wording_)

## What is the current behaviour?

If a user who isn't registered for an event tries to attend the event the system states that the card number does not exist. The card number might actually exist for a user, but the real thing here is that the user isn't actually signed up for the event.

## How do you reproduce this problem? 

Try to attend an event you're not signed up for and the resulting error message will be incorrect.

## Other information

This will most likely happen both when trying to sign up with username and by providing an RFID, based on the code [here](https://github.com/dotkom/onlineweb4/blob/develop/apps/events/views.py#L492-L505).

sklirg:
This also happens if no event ID is supplied when POSTing to the API.

PR #2152, opened 2018-10-17
Fix problem with regme saying user does not exists when it's not the case

tokongs:

- QA / Code Review

## Code Checklist

## Description of changes
Added a check to see if user exists in database when the user is not found as an attendee at event registration. 
FIxes issue #2052


Aborysa:
Looks good, just remember to add a case for this in the regme app before merging this 👍

apps/events/views.py
```diff
@@ -504,20 +504,32 @@ def post(self, request, format=None):
...
+                return Response({'message': (user[0].get_full_name() +
+                                '            er ikke meldt på dette arangementet.'),
+                                'attend_status': 55},
+                                status=status.HTTP_403_FORBIDDEN)
 
         # All is clear, attendee is attended
-        return Response({'message': (attendee.user.get_full_name() + ' er registrert som deltaker. Velkommen!'),
-                         'attend_status': 10, 'attendee': attendee.id}, status=status.HTTP_200_OK)
+        return Response({'message': (user.get_full_name() + ' er registrert som deltaker. Velkommen!'),
+                         'attend_status': 10, 'attendee': 2}, status=status.HTTP_200_OK)
```
sklirg: Why `2`?

tokongs: That was me testing something and forgetting to remove it. Will make a commit later today.

sklirg:
Furthermore, it looks like some of the test cases breaks with this change.

Aborysa:
Regme isn't worked on currently and probably won't be worked on for a long time.

PR #2156, opened 2018-10-24
Upgrade/django2

oleast:

- QA / Code Review

## Code Checklist

## Breaking Changes

**These changes are not backwards compatible**

- All model ForeignKeys are from now on required to define the `on_delete` keyword argument.

### Currently done:

- Update all existing ForeignKey fields to have `on_delete=models.CASCADE`. This was previously used as a default value by Django 1.x, but it has to be provided for all ForeignKey fields in Django 2.x.
- Django usage of `django.core.urlresolvers` to new package `django.urls`.
- Refactor all usage of `models.permalink` decorators in favor of `urls.reverse`.
- Import admin without `include() function`.
- Requirements:
  - Redwine: updated to Django 2 in a separate repo, updated commit reference.
  - Feedme: Has not been used for over 2 years. Removed from requirements and installed apps. R.I.P
  - Django Wiki: Updated to version 0.4.1 with support for Django 2.x
  - Django-dynamic-fixture: Updated to version 2 with support for Django 2.x
- Upgrade Python version to 3.5
- replace usage of the user.is_authenticated() method with user.is_authenticated property
- Force http403 middleware to work with new MIDDLEWARES system (from Django 1.10).
- Replace Widget._format_value with Widget.format_value
- Add renderer kwarg to widget renders
- Add timezone to test cases giving off warnings for naive datetimes.
- Make main url config work with Django 2 by setting app_name when the app uses namespace
- Change webshop CommaSeparatedIntegerField to Charfield with a validator
- _Add on_delete=models.CASCADE to __ALL__ required historic migrations_

- **Upgrade to Django version 2.0.9**

### ~Remaining work:~

- ~Add `on_delete=models.CASCADE` to historic migrations. This will only need to be done for all migrations made previous to Django 1.9 if we are lucky. [Details](https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.ForeignKey)~
- ~Change usage of 'namespaces' and 'app_name' in urls for Django 2 support. [Details](https://docs.djangoproject.com/en/dev/releases/1.9/#passing-a-3-tuple-or-an-app-name-to-include)~
- ~Finally, update the Django version in requirements.txt to `django==2.0.9`. (Intermediary before jumping to 2.1.x)~

### Other

- Make sure imports are sorted correctly
- Make sure tests work.


sklirg:
If upgrading the python version, make sure we run this version in
production as well.

oleast:
There has actually been a mismatch of python versions for a _long_ time now. The python version in production has been 3.5 since we upgraded to Debian 9.

The version in Docker was set recently. It was at `python:3` for a long time, which has been interpreted as the current release. The version was fixed to 3.4 when current resolved to 3.7, which broke with our current config.

requirements.txt
```diff
@@ -26,18 +26,17 @@ git+https://github.com/dotkom/django-guardian.git@1b184d5377200ab4ad59b343ae8af9
 django-taggit==0.22.2           # Generic multi-model tagging library
 django-taggit-serializer==0.1.5 # REST Framework serializers for Django-taggit
 APScheduler==3.5.1              # Scheduler
-feedme==1.9.5
-git+https://github.com/dotKom/redwine.git@0690807537fcbeeacde9c40b0e45549e0d216a96#egg=redwine==1.2.4.1
+git+https://github.com/dotkom/redwine.git@c26e12003a26bd4ef15c10d55358c24f96b3d132#egg=redwine
```
duvholt: A version should be provided otherwise I think pip will never upgrade the package once downloaded.
I think the previous version was fake since I (or whoever upgraded it last time) was too lazy to update the redwine version.

middleware/http.py
```diff
@@ -4,18 +4,24 @@
...
-
 """
 # Middleware to allow the display of a 403.html template when a
 # 403 error is raised.
 """
 
 
+try:
+    from django.utils.deprecation import MiddlewareMixin
+except ImportError:
```
duvholt: This shouldn't be necessary since we don't support multiple Django versions.

requirements.txt
oleast: Defined a version, and made a release of Redwine as well, just for the fun of it.

middleware/http.py
oleast: I'm a bit unsure of how to do it without doing this.

I guess we would have to rewrite the middleware to fit the new type?
https://docs.djangoproject.com/en/2.1/topics/http/middleware/#upgrading-middleware

duvholt: Rewriting it is probably a good idea, but what I actually meant was that the `try import-here except ImportError` code shouldn't be necessary. 
This kind of code is usually used in libraries that need to support multiple versions of Django and therefore have to provide a fallback in case the import is not included. 
Since we know that the `MiddlewareMixin` will be there we don't have to wrap it in `try except`.

oleast: Ah, I misunderstood a bit there, should be fixed now.

oleast: Would it be OK to just go for this solution here, and set up an issue for refactoring it?

duvholt: Absolutely! This PR is already big enough.

PR #2164, opened 2018-10-31
Add contribution app to api

martinlunde:
Feature request!
- Fetches dotkoms repositories from github (<2 years recent activity) each day and save these to database to avoid reaching the github rate limit.
- Opened api endpoint for OWF to get repository data

Can hopefully be used to implement the "contribution page" in OWF as mentioned earlier.
https://github.com/dotkom/onlineweb-frontend/tree/feat/contributions

## Code Checklist


apps/contribution/mommy.py
```diff
@@ -0,0 +1,97 @@
...
+                    type=language,
+                    size=(int(repo_languages[language])),
+                    repository=stored_repo
+                )
+                new_language.save()
+
+    @staticmethod
+    def new_repository(new_repo, new_languages):
+        # Filter out repositories with inactivity
+        if new_repo.updated_at.year > timezone.now().year - 2:
```
oleast: Wondering a little bit about the intention here.

Is it to get the repositories updated in the last two years from this month (eg. Nov 1st, 2016 -> Nov 1st, 2018), or at all those two years (eg Jan 1st, 2016 -> Nov 1st, 2018). Because this would do the latter one?

If so, would it not be kinda strange to suddenly drop a whole year of repositories when we reach a new year?

martinlunde: Indeed! Made a new commit resolving the problem and subtracting 2 years of the whole datetime object instead.

martinlunde:
Should be good now! Just the weird isort thing breaking the CI.

oleast:
Looks good to me, I'll just go ahead and merge it 😄

PR #2175, opened 2018-11-06
Fix broken flake 8 versioning

oleast:
… tox for linting locally

- QA / Code Review

## Code Checklist

## Description of changes

- Separate requirements for linting,
- upgrade python version to 3.5
- use tox for linting locally


Makefile
```diff
@@ -64,8 +64,7 @@ lint-only: lint-backend lint-frontend
 test-only: test-backend test-frontend
 
 lint-backend:
-	@docker-compose run --rm $(BACKEND_SERVICE_NAME) flake8 $(BACKEND_LINT_FOLDERS)
-	@docker-compose run --rm $(BACKEND_SERVICE_NAME) isort -c -rc $(BACKEND_LINT_FOLDERS)
+	@docker-compose run --rm $(BACKEND_SERVICE_NAME) tox -e flake8 -e isort $(BACKEND_LINT_FOLDERS)
```
sklirg: Tox specifies these folders manually, so they can be removed (`$(BACKEND_LINT_FOLDERS)`)

tox.ini
```diff
@@ -5,9 +5,8 @@ skipsdist = True
 [testenv]
 basepython = python3
 deps =
-    isort: isort
-    flake8: flake8
-            pyflakes
+    -r{toxinidir}/requirements.txt
```
sklirg: Do we need these requirements for linting? If we can skip installing all of Django and its requirements (and our 3rd party apps etc.) we can probably save quite some time.

PR #2187, opened 2018-11-27
Add filter for if a user is an attendee for an event

oleast:

- QA / Code Review

## Code Checklist

## Description:

- Add a filter which lets a logged in user filter events they are attendees on.
If the user is not logged in, all events are filtered away.

### DIFF
--- a/apps/events/filters.py
+++ b/apps/events/filters.py
@@ -1,7 +1,9 @@
+import logging
+
 import django_filters
 from django_filters.filters import Lookup
 
-from apps.events.models import Event
+from apps.events.models import AttendanceEvent, Attendee, Event
 
 
 class ListFilter(django_filters.Filter):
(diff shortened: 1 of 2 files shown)


apps/events/filters.py
```diff
@@ -17,8 +17,25 @@ class EventDateFilter(django_filters.FilterSet):
...
+        """ User cannot attend events if they are not logged in """
+        if not self.request.user.is_authenticated:
+            return queryset.none()
+
+        user_attendees = Attendee.objects.filter(user=self.request.user)
+        attending_events = [attendee.event.pk for attendee in user_attendees]
+        if value:
+            return queryset.filter(pk__in=attending_events)
+        else:
+            return queryset.exclude(pk__in=attending_events)
```
torjusti: I think we should not differ between not providing this and providing false. That is, to make this return all rows in the case where value is not provided.

apps/events/filters.py
```diff
@@ -17,8 +17,25 @@ class EventDateFilter(django_filters.FilterSet):
     event_end__gte = django_filters.DateTimeFilter(field_name='event_end', lookup_expr='gte')
     event_end__lte = django_filters.DateTimeFilter(field_name='event_end', lookup_expr='lte')
     attendance_event__isnull = django_filters.BooleanFilter(field_name='attendance_event', lookup_expr='isnull')
+    is_attendee = django_filters.BooleanFilter(field_name='attendance_event', method='filter_is_attendee')
     event_type = ListFilter()
 
+    def filter_is_attendee(self, queryset, name, value):
+        """
+        Filter events on if a user can attend them or not.
```
torjusti: Should not this say *Filter events based on if a user is attending them or not.*?

oleast: Yes, that is correct

apps/events/filters.py
oleast: I agree

oleast:
@torjusti How about now?

PR #2188, opened 2018-11-27
Fix profile privacy put request to correctly save a users privacy

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

Correct the updating method for user privacy settings to make it save that users privacy, and use a PUT request.

Companion PR of https://github.com/dotkom/onlineweb-frontend/pull/150

### DIFF
--- a/apps/profiles/views.py
+++ b/apps/profiles/views.py
@@ -572,10 +572,13 @@ class PersonalPrivacyView(viewsets.ViewSet):
         serializer = PrivacySerializer(privacy)
         return response.Response(serializer.data)
 
-    def update(self, request, pk=None):
+    def put(self, request, pk=None):
         privacy = Privacy.objects.get(user=request.user)
         serializer = PrivacySerializer(privacy, data=request.data)
-        return response.Response(serializer.data)
+        if serializer.is_valid():
+            serializer.save(user=request.user)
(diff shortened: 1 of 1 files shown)


apps/profiles/views.py
```diff
@@ -572,10 +572,13 @@ def list(self, request, format=None):
         serializer = PrivacySerializer(privacy)
         return response.Response(serializer.data)
 
-    def update(self, request, pk=None):
+    def put(self, request, pk=None):
         privacy = Privacy.objects.get(user=request.user)
         serializer = PrivacySerializer(privacy, data=request.data)
-        return response.Response(serializer.data)
+        if serializer.is_valid():
```
sklirg: If you use add the kwarg `raise_exception=True` here it will automatically return a HTTP 400 BAD REQUEST if the serializer validation fails, rather than a 200 OK for an invalid/non-chaning PUT request.

https://www.django-rest-framework.org/api-guide/serializers/#raising-an-exception-on-invalid-data

oleast: Ah, I did not know that.
Should I still return something if it is not valid, or just count on the exception handling in DRF to take care of it in all cases?

sklirg: It'll raise an exception so you won't need to return anything, no

PR #2190, opened 2019-01-23
Serialize RFID field and enable PUT to profile API

torjusti:
Serialize the RFID field and enable PUT requests to the profile API.
The intention is to allow updating RFIS values, so that #1635 can be fixed in OWF.

- QA / Code Review

## Code Checklist

### DIFF
--- a/apps/profiles/views.py
+++ b/apps/profiles/views.py
@@ -587,3 +587,10 @@ class ProfileViewSet(viewsets.ViewSet):
         user = request.user
         serializer = ProfileSerializer(user)
         return response.Response(serializer.data)
+
+    def put(self, request, pk=None):
+        user = request.user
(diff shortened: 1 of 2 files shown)


apps/profiles/views.py
```diff
@@ -587,3 +587,10 @@ def list(self, request, format=None):
         user = request.user
         serializer = ProfileSerializer(user)
         return response.Response(serializer.data)
+
+    def put(self, request, pk=None):
+        user = request.user
+        serializer = ProfileSerializer(user, data=request.data, partial=True)
+        if serializer.is_valid(raise_exception=True):
+            serializer.save(user=request.user)
```
oleast: `user` is already defined as a variable, could use that instead of `request.user`?

PR #2200, opened 2019-02-13
checkbox for being anon for contact form

kharann:

- QA / Code Review

## Code Checklist

- [X] The code is ready to be merged

- [X] The changes are backwards compatible

## Description of changes
Added an optional checkbox for being anon. This checkbox will disable name and email form fields. I  also changed name and email fields to be required by default. Things where being hard by the fact i couldn't use functions inside the index, hence i putted the whole function inside. Sadly this is kind of a hacky way.

## Screenshot(s) if appropriate provide screenshots (before and after) if doing design changes -->

isanon1
isanon2

### DIFF
--- a/apps/contact/forms.py
+++ b/apps/contact/forms.py
@@ -8,8 +8,18 @@ class ContactForm(forms.Form):
                     )
     contact_receiver = forms.ChoiceField(required=True, label="Hvem ønsker du å kontakte?", choices=mail_choices)
     contact_checkbox = forms.BooleanField(required=False)
-    contact_name = forms.CharField(required=True, widget=forms.TextInput({"placeholder": "Navn"}))
-    contact_email = forms.EmailField(required=True, widget=forms.TextInput({"placeholder": "Epostadresse"}))
+    contact_name = forms.CharField(required=False, widget=forms.TextInput({"placeholder": "Navn"}))
+    contact_email = forms.EmailField(required=False, widget=forms.TextInput({"placeholder": "Epostadresse"}))
     content = forms.CharField(required=True, widget=forms.Textarea({"placeholder": "Din melding"}))
     captcha = ReCaptchaField(error_messages={'required': ('Du klarte ikke captchaen! Er du en bot?'),
                                              'invalid': ('Du klarte ikke captchaen! Er du en bot?')})
+
+    def clean(self):
+        name = self.cleaned_data['contact_name']
+        is_anon = self.cleaned_data['contact_checkbox']
+        email = self.cleaned_data['contact_email']
+
+        if not (name and email) and not is_anon:
+            raise forms.ValidationError(
+                "Du må skrive ditt navn og mail hvis du ikke skal være anonym!"
+            )
--- a/apps/contact/forms.py
+++ b/apps/contact/forms.py
@@ -10,7 +10,7 @@ class ContactForm(forms.Form):
     contact_checkbox = forms.BooleanField(required=False)
     contact_name = forms.CharField(required=False, widget=forms.TextInput({"placeholder": "Navn",
                                                                            "required": True}))
-    contact_email = forms.EmailField(required=False, widget=forms.TextInput({"placeholder": "Epostadresse",
+    contact_email = forms.EmailField(required=False, widget=forms.EmailInput({"placeholder": "Epostadresse",
                                                                              "required": True}))
     content = forms.CharField(required=True, widget=forms.Textarea({"placeholder": "Din melding"}))
     captcha = ReCaptchaField(error_messages={'required': ('Du klarte ikke captchaen! Er du en bot?'),
@@ -22,6 +22,6 @@ class ContactForm(forms.Form):
         email = self.cleaned_data.get('contact_email')
 
         if not (name and email) and not is_anon:
-            raise forms.ValidationError(
-                "Du må skrive ditt navn og mail hvis du ikke skal være anonym!"
-            )
+            error_msg = "Must be filled"
+            self.add_error('contact_name', error_msg)
+            self.add_error('contact_email', error_msg)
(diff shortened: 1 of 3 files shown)


kharann:
> Works and looks good on the front-end part, but it will not work in the back-end.
> 
> Since `contact_name` and `contact_email` are marked as required in the Django form, the validation will always fail when they are not provided, no matter what has been done in the front-end (Since you can't trust front-end code as it can be manipulated by the user).
> 
> To fix this, you will have to mark the fields as not required, and validate the field with some extra logic in the ContactForm class.
> 
> This can be done using the `clean` method of the form ([Documentation](https://docs.djangoproject.com/en/dev/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other)). You'll have to check if the checkbox is checked, and if it is not; validate the name and email, and raise a descriptive `ValidationError` if they are not present.

I believe this should fix the issue. I tested it by trying to send a mail without name and mail filled out the message wouldnt go through. When i checked the box and did the same the message go through.

If the name and email fields are empty and the checkbox is unchecked, you will end up with the error "Meldingen ble ikke sendt. Prøv igjen eller send mail direkte til dotkom ". But thats mostly because of the error message in contact/views.py

oleast:
I'm being a bit picky here, but I think it is important to get it right.

You'll get a server error if the 'contact_email' field is not an actual email.
Seems that since the field is not valid; it is not included in the 'cleaned_fields' dict.
That makes it result in an error (I think).

The 'contact_email' and 'contact_name' should be marked as required on the front-end when the user does not want to be anonymous.
This would make things a little clearer for the user.

I think we should also try to find a way to get the error message to display correctly, but it is not as important if the other features are implemented.

kharann:
> I'm being a bit picky here, but I think it is important to get it right.
> 
> You'll get a server error if the `contact_email` field is not an actual email.
> Seems that since the field is not valid; it is not included in the `cleaned_data` dict.
> That makes it result in an error (I think).
> 
> The `contact_email` and `contact_name` should be marked as required on the front-end when the user does not want to be anonymous.
> This would make things a little clearer for the user.
> 
> I think we should also try to find a way to get the error message to display correctly, but it is not as important if the other features are implemented.

It seems like changing the cleaned_data['contact_mail'] to cleaned_data.get('contact_mail'). Also, since the frontend is generated from the ContactForm class, which is marked with required=False (which we have to since if else we end up with the issue at the start when required is True). We could maaybe use javascript to make it so the field is only required at the front end since i cant directly change the html? Any ideas?

oleast:
> 
> It seems like changing the cleaned_data['contact_mail'] to cleaned_data.get('contact_mail'). Also, since the frontend is generated from the ContactForm class, which is marked with required=False (which we have to since if else we end up with the issue at the start when required is True). We could maaybe use javascript to make it so the field is only required at the front end since i cant directly change the html? Any ideas?

It may be possible to use set the required attribute on the fields without defining them in the Django FormField. Something like [https://docs.djangoproject.com/en/2.1/ref/forms/widgets/#django.forms.Widget.attrs](Widgets with attrs)?

This will make them required at the front-end, but not in the back-end validation of the form data. It should also work when the fields are `disabled`, as that should override the `required` attribute.

kharann:
Should be resolved now!

oleast:
Great work, I think we are nearing completion soon!

We still have the problem of the email field not being validated as an email on the front-end. The `type` attribute is still set to `text`.
This can be fixed by setting the widget to a `forms.EmailInput`.

assets/contact/index.js
```diff
@@ -1 +1,13 @@
 import './less/contact.less';
+
+const checkbox = document.getElementById('id_contact_checkbox');
+const contactName = document.getElementById('id_contact_name');
+const contactMail = document.getElementById('id_contact_email');
+
+checkbox.addEventListener('change', () => {
+  if (checkbox.checked) {
```
oleast: In this case, we don't really need the if-statement at all.

We are already using the `checked` value of the checkbox to determine the `disabled` attribute.
That means we can just directly set it based on the value:
``` javascript
contactMail.disabled = contactName.disabled = checkbox.checked;
```
Or maybe a bit clearer on this form:

``` javascript
contactMail.disabled = checkbox.checked;
contactName.disabled = checkbox.checked;
```

assets/contact/index.js
```diff
@@ -1 +1,13 @@
...
+    contactMail.disabled = contactName.disabled = true;
+  } else {
+    contactMail.disabled = contactName.disabled = false;
```
oleast: In this case, we don't really need the if-statement at all.

We are already using the `checked` value of the checkbox to determine the `disabled` attribute.
That means we can just directly set it based on the value:
``` javascript
contactMail.disabled = contactName.disabled = checkbox.checked;
```
Or maybe a bit clearer on this form:

``` javascript
contactMail.disabled = checkbox.checked;
contactName.disabled = checkbox.checked;
```

apps/contact/forms.py
```diff
@@ -7,8 +7,21 @@ class ContactForm(forms.Form):
...
+
+    def clean(self):
+        name = self.cleaned_data.get('contact_name')
+        is_anon = self.cleaned_data.get('contact_checkbox')
+        email = self.cleaned_data.get('contact_email')
+
+        if not (name and email) and not is_anon:
+            raise forms.ValidationError(
+                "Du må skrive ditt navn og mail hvis du ikke skal være anonym!"
+            )
```
sklirg: I'd prefer if the error were attached to fields. If I'm not mistaken completely, this is accomplished by using `ValidationError({'field_name': 'Validation error text'})`. 

However, this resource shows the feature in a simpler way: https://docs.djangoproject.com/en/2.1/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other

If you use `self.add_error(field, text)` you can add the error to a specific field.

Could you look into that? 😄 

And conditionally check each field, obviously, so that you get one error per field instead of one error saying that one of the fields are missing.

Issue #2182, opened 2018-11-21
Remove list of email addresses from mail-participants page.


aslakhol:
## What kind of an issue is this?

- Enhancement
- GDPR

## What is the current behaviour?
On the mail-participants page for an event we list all the email adresses of attendees. This leads to issues in events with a lot of attendees since the list becomes so long that it is no longer obvious that we have a form for sending email at the bottom. 

There is also no reason for this page to display the email adresses and we should be limiting this information where we can, to be more compliant with behaviour. 

## What is the expected behaviour?
The page should not display a list of all email, it should only display the form for sending email. 

We could consider to list attendees here, but I think that would be unnecessary as they are listed in other places. 

## How do you reproduce this problem? 

Go to mail-participants page for an event.

https://online.ntnu.no/events/mail-participants/597

oleast:
Fixed by #2204

PR #2204, opened 2019-02-20
changed list of mail addresses to amount of participants

kharann:

- QA / Code Review

## Code Checklist

## Description of changes
Issue: Remove list of email addresses from mail-participants page. Easy GDPR
#2182 
Changed the loong list of mails to just number of the following ones instead.
whatsupguys

### DIFF
--- a/templates/events/mail_participants.html
+++ b/templates/events/mail_participants.html
@@ -1,8 +1,14 @@
 {% extends 'base.html' %}
+{% load render_bundle from webpack_loader %}
 
 {% block title %}
 Send epost til påmeldte - Online
 {% endblock title %}
+
+{% block js %}
+	{{ block.super }}
+	{% render_bundle 'eventsDetails' 'js' %}
+{% endblock %}
 
 {% block content %}
 <section>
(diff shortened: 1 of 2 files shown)


templates/events/mail_participants.html
```diff
@@ -14,37 +14,15 @@ <h2>Send epost til påmeldte for  «{{ event.title }}»</h2>
                 </div>
             </div>
         </div>
-        <div class="row">
+        <div class="row" style="font-size: 18px">
```
sklirg: I'd rather see this done through CSS than inline styling.

kharann: I was wondering how i should do this since i couldnt find a the index.js/css file to the html. Should i create a new bundle in assets/events with a index.js which imports the css document or done some other way?

kharann: put the css inside of the assets/events/details/ less document since i believe making a new bundle just for these 3 lines so i included it inside an existing css document

kharann: thought it was fine since since they used inline in. But it makes sense that we shouldnt use inline at all i guess
```html
<div class="row" style="margin-top:30px">
```

PR #2220, opened 2019-03-08
Add option to hide event

### LINKED ISSUE #2206: Favor hiding events to deleting them.

- Feature request

## What is the expected behavior?

Deleting events may cause a lot of unnecessary consequences, like deleting payments and generally losing history of what has been done.

It should therefore be possible to hide an event with a boolean setting on the event.
Hiding it should be used in most cases, but there should be a dashboard override for deleting in case it is strictly necessary.

## What is the current behavior?

The only way to make an event not visible anymore is to delete it from the database.

henrikhorluck:
- Hides on frontpage
- Hides on /events
- Hides on detailView

#### TODO: 

- [x] Add tests
- [x] Add option to hide event in Dashboard

Fixes #2206 
- QA / Code Review

## Code Checklist


apps/events/models.py
```diff
@@ -169,7 +173,7 @@ def can_display(self, user):
             return True
         if not user:
             return False
-        return restriction.has_access(user)
+        return restriction.has_access(user) if self.visible else False
```
oleast: Would this not still return `True` if the event has no GroupRestriction?

henrikhorluck: Yes, because of `if not restriction: return True` earlier in the function

oleast: If I'm not completely wrong `visible` should have higher precedence than a GroupRestriction?

In this case, all events will be visible if they don't have a GroupRestriction, which is almost all events.

henrikhorluck: Yes, you are correct, fixed it now.

apps/events/models.py
```diff
@@ -114,6 +114,10 @@ class Event(models.Model):
     """Event type. Used mainly for filtering"""
     organizer = models.ForeignKey(Group, verbose_name=_('arrangør'), blank=True, null=True, on_delete=SET_NULL)
     """Committee responsible for organizing the event"""
+    visible = models.BooleanField(
+        _('Vis arrangementet utenfor Dashboard/admin-panelet'),
+        default=True,
+        help_text='Denne brukes primært for å skjule eksisterende eventer.')
```
oleast: I think wording it as 'arrangementer' would be more correct than 'eventer'

henrikhorluck: I agree, fixed

PR #2221, opened 2019-03-08
Restrict attendee endpoint to only list attendees for event user can …

oleast:
…change

- QA / Code Review

## Code Checklist

Applications depending on getting attendees without restrictions may break

## Description of changes

Restrict list of attendees to only show attendees for the events a user has access to change.

### DIFF
--- a/apps/events/filters.py
+++ b/apps/events/filters.py
@@ -1,5 +1,6 @@
 import django_filters
 from django_filters.filters import BaseInFilter, NumberFilter
+from guardian.shortcuts import get_objects_for_user
 
 from apps.events.models import Attendee, Event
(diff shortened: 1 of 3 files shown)


oleast:
Should we maybe get all if the user is a superuser?

oleast:
Update:
- Adds filter for event API view:
```
?can_change=(True|False)
```

- Restrict attendance registration view to require permission to change event as well

oleast:
**Requires changes to tests**

oleast:
Changes have successfully been ported to OIDC and have been tested to work with Regme.

I think this is ready to be merged on Wednesday?

Issue #2183, opened 2018-11-21
Add ability to include images in emails sent from OW.


aslakhol:
## What kind of an issue is this?

- Feature request

## What is the expected behaviour?
We would like to be able to include images in email sent to participants of events.

## What is the current behaviour?
You can't and as such some emails are sent manually, not using the system.

kharann:
Is it ok to have the mail as an attachment fine? Or do you wish the image to be shown in the mail

oleast:
Fixed by #2228

PR #2228, opened 2019-03-19
Attach image to mail

kharann:
This pullrequest should implement Issue #2183. With this pullrequest i believe you should be able to add images as attachments and it should be attached to the mail. You're also able to choose multiple attachments. Not sure if its gonna be embeded or attached since i wasnt able to view the physical mail in the terminal, however i can see the image in raw text format at least. I havent added any max size or anything since idk what the limit should be
## Code Checklist

## Description of changes
Adds an optional file upload input in the form for sending mail to all event participants such that you can send multiple images. 

yahoo


henrikhorluck:
Would like if you added tests for this :)

PR #2246, opened 2019-04-25
Fix #2245, using email-object instead of email-address

### LINKED ISSUE #2245: Membership application uses email object instead of email-address
Same as #2219

https://sentry.io/organizations/dotkom/issues/890887549/?project=204971&referrer=github_plugin

```
TypeError: object of type 'Email' has no len()
(16 additional frame(s) were not displayed)
...
  File "guardian/decorators.py", line 126, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "apps/approval/dashboard/views.py", line 105, in approve_application
    app.save()
  File "apps/approval/signals.py", line 50, in notify_membership_applicant_handler
    send_approval_status_update(instance)
  File "apps/approval/tasks.py", line 45, in send_approval_status_update
    [approval.applicant.get_email()],

TypeError: object of type 'Email' has no len()
```

henrikhorluck:
Close #2245

### DIFF
--- a/apps/approval/signals.py
+++ b/apps/approval/signals.py
@@ -55,4 +55,4 @@ def notify_new_committee_application(sender, instance, created, **kwargs):
     if created:
         send_committee_application_notification(instance, [settings.EMAIL_OPPTAK], link_to_admin=True)
         if settings.APPROVAL_SETTINGS.get('SEND_COMMITTEEAPPLICATION_APPLICANT_EMAIL', False):
-            send_committee_application_notification(instance, [instance.get_email()], link_to_admin=False)
+            send_committee_application_notification(instance, [instance.get_email().email], link_to_admin=False)
--- a/apps/approval/tasks.py
+++ b/apps/approval/tasks.py
@@ -42,7 +42,7 @@ def send_approval_status_update(approval):
         EmailMessage("Soknad om medlemskap i Online er vurdert",
                      message,
                      settings.DEFAULT_FROM_EMAIL,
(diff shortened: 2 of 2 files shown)


apps/approval/signals.py
```diff
@@ -55,4 +55,4 @@ def notify_new_committee_application(sender, instance, created, **kwargs):
     if created:
         send_committee_application_notification(instance, [settings.EMAIL_OPPTAK], link_to_admin=True)
         if settings.APPROVAL_SETTINGS.get('SEND_COMMITTEEAPPLICATION_APPLICANT_EMAIL', False):
-            send_committee_application_notification(instance, [instance.get_email()], link_to_admin=False)
+            send_committee_application_notification(instance, [instance.get_email().email], link_to_admin=False)
```
henrikhorluck: This is not the same issue, but as far as I can see in the docs, django's `send_email()` (that `send_committee_application_notification` invokes) uses a list of strings as input, not email-objects, and thus this fixes that as well

oleast:
This seems to be a quite regular problem for us, it has happened before.

I would guess the naming confuses people, it at the very least seems odd to me.
I think we should rename the emails to something a bit less confusing?

henrikhorluck:
Something like `get_email_object`? I mean what else is in the email?

henrikhorluck:
Okay so I checked the model for the email, and it has stuff like wether it is verified, and which user it links to, which is actually used in some places, event though most use the `.email`-option. I do not think it is necessary to change, as of now

PR #2260, opened 2019-05-08
Upgrade linting

henrikhorluck:
Upgrade linting tools, fix linting
- Fix over-indenting
- Add ignore rules to unused imports
- Remove duplicate imports
- Rename and refactor test method
- Make regex-strings raw-string-literals


henrikhorluck:
Already working on it on a local branch, although there are some breaking changes, e.g. in djangorestframework

PR #2258, opened 2019-05-07
Create API for attending events

oleast:
## What kind of a pull request is this?

- QA / Code Review

## Code Checklist

- [x] The code follows dotkom code style 
- [x] The code passes the defined tests
- [x] I have added tests for the code I added
- [ ] I have provided documentation for the code I added
- [x] The code is ready to be merged

## (Possible) Breaking Changes

- [x] The changes are backwards compatible

## Description of changes

- Create new API endpoint for attending and un\-attending events
- Create API endpoint for viewing user related attendance-event
- Add tests for attendance endpoint
- Add an API filter for if a user can attend the event
- Refactor payment complexity for attendee
- Implement recaptcha for event attendance
- Support extras in attendance event serialization, and for attendance

## Other changes
- Add generic testcase for OIDC API testing
- Add payment price to generated Payment during event testing (?)
- Implement recaptcha as a reusable serializer field
- Fix `attendance_event#is_attendee` returning a `QuerySet`, not a `boolean`

## Notes

- The functionality of payments has **not** been implemented as an API yet. This means it will not be possible to pay for events yet. Attendees can however still see if they have paid or not.

- There could always be more tests, I guess. A lot of the testing is covered in the model tests, and the combined view tests for the old events.

## Screenshot(s) if appropriate

I suspect screenshots are not needed for API-changes 🤷‍♂

PR #2261, opened 2019-05-09
Complete signup process through API

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

Relies on things done in #2258 

Adds full support for signing up a user. Includes:
- Captcha for singup
- Setting related email object and primary email.
- Sending verification email to activate the user.
- Updating user data
- Creating, updating, deleting, and viewing user emails
- Creating, updating, deleting, and viewing user positions
- Viewing user special positions


apps/authentication/api/views.py
```diff
@@ -1,22 +1,110 @@
...
+        return super().get_serializer_class()
+
+    def perform_destroy(self, instance: User):
+        """
+        Destroying a user should delete all personal data, and mark the user as inactive.
+        All fields that can be wiped should be wiped, and other fields should be made anonymous.
+        Deleting the user object itself should not be done, as relations like attendance should be preserved.
+        """
+        instance.active = False
+        instance.save()
```
henrikhorluck: Is this method actually doing what the docs are saying?

sklirg: I don't think the API view should handle this. If this is to be done, it should be done as a post_save trigger. And that is outside the scope of this PR (imo).

oleast: May have been the wrong place to do this, at least the wrong place to write the thoughts.

I did not implement the actual deletion/scrubbing of user data because we have not currently discussed how we should actually do this. We have a lot of user-related data, and we have to find out what we want to delete and what we want to make anonymous.

I think this PR is already a bit too large, and this should be considered a thing to implement in the future.

@sklirg That could be done, do you have any thoughts on how that trigger/signal should be triggered? I suspect you would have to mark the user as deleted or something to make the signal only delete the user-info in specific cases?
The `active` field does not seem like the right thing to use in that case, as I would think that could be set to `False` in other situations.

Did I misunderstand, or is the API-view also the wrong place to trigger account deletion?

oleast:
Fixed the test error, which somehow came from duplicate instances of the same RFID on a user during testing for an unrelated part of the app :woman_shrugging: 

Is there anything blocking this from being implemented now? It would be quite nice to have a way to create users and such.

apps/authentication/api/views.py
sklirg: You can override the delete method on the model.

What I meant is that the API is not the place to implement account deletion functionality. That should be done on the model. The API should in turn call that method for deleting the object, which will use the method from the model.

oleast: Ah, that sound more like the right way to do it.

I've updated the PR and removed the `destroy` option from the viewset.
It can be re-introduced if user-scrubbing is implemented.

PR #2264, opened 2019-05-10
Upgrade dependencies

henrikhorluck:

- QA / Code Review

## Code Checklist

## Description of changes
Start of dependency upgrades, mostly changes without breaking changes so far, but will add and fix the rest soonTM


assets/common/utils/MicroEvent.js
```diff
@@ -13,7 +13,8 @@ const MicroEvent = () => {
      * @param {*} [payload] An optional event payload
      */
     fire(event, payload) {
-      if (debug) console.log(`Firing event: ${event}`, 'Payload: ', payload);
+        // eslint-disable-next-line no-console
```
oleast: Why are these indented more than the line following them?

requirements.txt
```diff
@@ -1,62 +1,56 @@
...
 django-filter==2.1.0           # Filtering for DRF
 # python-memcached==1.57          # Memcache. Used by Mailinglists to fetch from Sympa.
 # Upstream is missing Python 3 patches
 git+https://github.com/JelteF/python-memcached@patch-1#egg=python-memcached==1.58
 markdown2==2.3.7                # textarea text formatting
 pytz                            # Timezone lib
-stripe==2.27.0                  # Stripe payment
-icalendar==4.0.1                # iCalendar generation
-google-api-python-client==1.6.5
+stripe==2.28.1                  # Stripe payment
```
oleast: I'm a bit unsure about this upgrade.
I originally built the new payments API with `2.27.0`, and everything was working great.
Tested it with `2.28.0` before setting up the PR, but random tests were failing because of API problems every time.

Maybe they fixed it with `2.28.1`, but I think it would be wise to look into.

requirements.txt
```diff
@@ -1,62 +1,56 @@
 Django==2.1.8
-Pillow==4.3.0                   # PIL fork (Python Imaging Library)
-djangorestframework==3.7.7      # API v1
+Pillow==6.0.0                   # PIL fork (Python Imaging Library)
```
oleast: We are upgrading two major versions here, any idea what has actually been deprecated?
I found little to no information in their releases on Github.

requirements.txt
```diff
@@ -1,62 +1,56 @@
...
-django-oauth-toolkit==1.0.0    # OAuth2 authentication support
+django-recaptcha==2.0.4         # Google ReCaptcha
+django-oauth-toolkit==1.2.0    # OAuth2 authentication support
 django-watson==1.5.2            # Indexed model search lib
-django-reversion==2.0.13        # Model version history with middleware hooks to all post_save signals
-django-guardian==1.5.0          # Per Object permissions framework
-django-taggit==0.23.0           # Generic multi-model tagging library
-django-taggit-serializer==0.1.5 # REST Framework serializers for Django-taggit
-APScheduler==3.5.1              # Scheduler
+django-reversion==3.0.4        # Model version history with middleware hooks to all post_save signals
```
oleast: I think it would be wise to be careful here.

From what I can see we've had problems with this before.
We cant see the version history of anything older than 2016 because of a change in this.

apps/events/views.py
```diff
@@ -630,3 +630,7 @@ def post(self, request, format=None):
         # All is clear, attendee is attended
         return Response({'message': (attendee.user.get_full_name() + ' er registrert som deltaker. Velkommen!'),
                          'attend_status': 10, 'attendee': attendee.id}, status=status.HTTP_200_OK)
+
+    @staticmethod
+    def get_extra_actions():
+        return []
```
oleast: Is this for some strange reason absolutely needed for DRF 3.9?

henrikhorluck: This has to do with DRF 3.8 adding the `get_extra_actions`-method to Views [(PR)](https://github.com/encode/django-rest-framework/pull/5705/files), [Changelog](https://www.django-rest-framework.org/community/release-notes/#38x-series), so what I think is happening here is that `AttendViewSet`, which inherits from `APIViewSet`, and not a generic viewset (viewsets.ViewSet), does not by default have the `get_extra_actions`-method on it, but we are for some reason using it as a "normal" viewset, _somewhere_, this _might_ be https://github.com/dotkom/onlineweb4/blob/358b15f1bc3ec25c8dd530b9186c6ab3a169beae/apps/events/urls.py#L38 and I am checking right now if adding `.as_view()` fixes the problem

henrikhorluck: Update: nope, gives `AttributeError: 'function' object has no attribute 'get_extra_actions'
`

requirements.txt
henrikhorluck: As far as I can see (or the IDEA can search) we only use PIL in https://github.com/dotkom/onlineweb4/blob/358b15f1bc3ec25c8dd530b9186c6ab3a169beae/apps/gallery/util.py#L10, and we use relatively few functions, that I do not think are affected by their deprecations https://pillow.readthedocs.io/en/stable/releasenotes/index.html

requirements.txt
henrikhorluck: The tests are failing locally, but that has to do with me not having a test-stripe-token, and they are working on CI, so it should be fine

assets/common/utils/MicroEvent.js
oleast: _These_ as in this and the one a few lines below as well

henrikhorluck: Indenting 2 spaces is hard with .editorconfig set to 4, okay?! :(

Issue #2240, opened 2019-04-17
[Before 14.09.19] Upgrade Stripe to comply with SCA


oleast:
## What kind of an issue is this?

- Feature request

## What needs to be done

Because of regulation changes, we have to upgrade Stripe before 14.09.2019

## Other information

Stripe information:
[About PSD2 and SCA](https://stripe.com/en-NO/payments/strong-customer-authentication?utm_medium=email&utm_campaign=SCA%20Migration%20Wave%203%20-%20Direct%20-%20EU%20EN&utm_content=SCA%20Migration%20Wave%203%20-%20Direct%20-%20EU%20EN+CID_4944d79d9d826ec0309f4fa468842efb&utm_source=&utm_term=Strong%20Customer%20Authentication%20SCA)
[Stripe Docs](https://stripe.com/docs/strong-customer-authentication?utm_medium=email&utm_campaign=SCA%20Migration%20Wave%203%20-%20Direct%20-%20EU%20EN&utm_content=SCA%20Migration%20Wave%203%20-%20Direct%20-%20EU%20EN+CID_4944d79d9d826ec0309f4fa468842efb&utm_source=&utm_term=Read%20the%20docs)

oleast:
Resolved

PR #2266, opened 2019-05-12
Create REST api for payments

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

Adds a REST API implementation for the current payment solutions.

- Event payments should be done and properly tested
- Transactions (Nibble/saldo) is implemented.
- Thinking about implementing it. Awaiting a bit on it.

Implements Stripe by using the Intents API instead of the Charges API.
Solves the back-end part of #2240


oleast:
*Note: Tests are currently failing because stripe token has not been set up in the test env. I'm a bit unsure if we should:
- Mock the Stripe API to always return a success somehow? (Would probably be hard)
- Ignore them in the CI, at least in most cases? (Would be annoying if you don't test well enough locally, which probably happens too often)
- Set up a [mock Stripe server](https://github.com/stripe/stripe-mock)? (Would make the tests a lot faster, they are currently annoyingly slow)
- Just deal with it and let the CI run tests against the Stripe API?

PR #2269, opened 2019-05-16
Fix renderer keyword in widget render from deps

oleast:
Upgrade deps to fix issues with renderer keyword in widget render methods.


sklirg:
Just a heads up - since you "faked" the version number to the sequential next version for the package, when that version gets released pip will think it is already installed since the version number matches. 
If I do things like this I usually add another number, like 1.9.3.1 or so to avoid this problem.

PR #2281, opened 2019-06-27
Create a somewhat working group system

### LINKED ISSUE #1209: Revamp how we handle committees

oleast:

- QA / Code Review

## Code Checklist

The changes don't _require_ changes, but data will have to be fixed by hand to make the group data correct, as the required data does not exist in the DB.

## Description of changes

I've been thinking for a while that it is sad that we cannot do the things we want because our group system is lacking vital features.

I made a little effort to bring out a basic group system that can be extended in the future, but will work more in line with how we actually use it.

The features are supposed to be fully backwards compatible, and should only work as an addition to the current group system.

This PR introduces three new models:
- OnlineGroup, a OneToOne related model to Django groups which holds actual information about groups in Online.
- GroupMember, a model relation between a user and an OnlineGroup.
- GroupRole, which relates a GroupMember to a set of roles they can have, like being the leader or treasurer of the group.

Fixes #1209 ? :sheep:


oleast:
I'm leaning towards just merging this now, any thoughts?

PR #2293, opened 2019-08-22
Fiken sale registration

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

- Adds automagic registration of sales with Stripe to Fiken


apps/payment/constants.py
```diff
@@ -0,0 +1,111 @@
...
+        HIGH_PURCHASE_OF_EMISSIONSTRADING_OR_GOLD_NONDEDUCTIBLE,
+    )
+    ALL_CHOICES = [(t, t) for t in ALL_TYPES]
+
+
+class FikenAccount:
+    ARRKOM = '1960:10003'
+    PROKOM = '1960:10002'
+    TRIKOM = '1960:10001'
+    FAGKOM = '1960:10004'
```
sklirg: Are these Fiken internal identifiers? Should they be stored as configuration parameters (á la Stripe keys per committee) instead of hard coded in the class?

apps/payment/constants.py
```diff
@@ -0,0 +1,111 @@
+# Make sure these exist in settings if they are to be used.
```
sklirg: I'd reword the comment to say "These settings are configured here: settings/file.py".

apps/payment/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+FIKEN_SALES_API_URL = f'{FIKEN_API_URL}/companies/{FIKEN_ORG}/sales'
+
+
+logger = logging.getLogger(__name__)
+
+
+@app.task(bind=True)
+def register_sale_with_fiken(_, sale_data: dict):
+    logger.info('Starting Fiken register')
+    if FIKEN_USER and FIKEN_PASSWORD and FIKEN_ORG:
```
sklirg: Add an `else` clause and log it

apps/payment/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+        response = requests.post(
+            url=FIKEN_SALES_API_URL,
+            auth=auth,
+            headers=headers,
+            json=sale_data,
+        )
+        if response.ok:
+            logger.info(f'Successfully registered sale {sale_data.get("identifier")} in Fiken')
+        else:
+            logger.warning(f'Failed at registering sale {sale_data.get("identifier")} in Fiken')
```
sklirg: Maybe log some of the response, e.g. the error message and the HTTP status code?

apps/payment/models.py
```diff
@@ -553,3 +580,106 @@ def _send_receipt(self, payment_date, subject, description, payment_id, items, t
...
+        max_length=20,
+        choices=StripeKey.ALL_CHOICES,
+        default=StripeKey.ARRKOM,
+    )
+    transaction_type = models.CharField(
+        'Transaction Type',
+        max_length=20,
+        choices=TransactionType.ALL_CHOICES,
+    )
+    original_amount = models.FloatField()
```
sklirg: Don't use floats for money. Especially if stored internally and is later used for calculations^1.

Here's a nice, short read: https://husobee.github.io/money/float/2016/09/23/never-use-floats-for-currency.html

I'd suggest to keep the "Stripe way", as in keeping the base "øre" amount ("100.50 kr" would be "10050 øre") and use that. That way you never have to deal with floating point arithmetic and its problems. If Fiken accepts only "kroner and øre" as a float (first off: WTF), convert it when posting to their API rather than storing this internally.

Why?
```python
>>> 0.1 + 0.2
0.30000000000000004
```
Source: https://docs.python.org/3.7/tutorial/floatingpoint.html

1) It might not be a problem here because it is only stored in this class to later be posted to Fiken, but better handle it correctly from the first time around IMO. Suddenly this is getting rewritten without people thinking about this and then you get the problems.

---

So I'd suggest creating a "to_fiken_api_format" function on the class which converts the model to a JSON representation that Fiken accepts, which converts the model class into whatever format Fiken expects.

apps/payment/signals.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@receiver(post_save, sender=FikenSale)
+def handle_fiken_sale_created(sender, instance: FikenSale, created=False, **kwargs):
+    if created:
+        logger.info(f'Registering {instance} with Fiken')
+        sale_data = FikenSaleSerializer(instance).data
+        register_sale_with_fiken.delay(sale_data=sale_data)
+    else:
+        logger.warning(f'{instance} updated after creation')
```
sklirg: Should this be denied? If so, do that at the model layer, rather than having out-of-sync items in OW4 and Fiken.

If this should be supported in the long run, disable it for now and tell someone to contact dotkom/bankom to fix it for now, and tell them that this functionality is not supported yet.

apps/payment/migrations/0031_fikensale_description.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('payment', '0030_fikensale'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='fikensale',
+            name='description',
+            field=models.CharField(default='TEST', max_length=200),
```
sklirg: This doesn't look useful :p I'd rather have an empty string than a "test" string.

apps/payment/models.py
```diff
@@ -354,6 +356,17 @@ def get_from_mail(self):
     def get_to_mail(self):
         return self.user.email
 
+    def _create_fiken_sale(self, amount: float):
+        transaction_type = TransactionType.EVENT
```
sklirg: Minor nitpick: Why is this assigned here and only used one other place? Why not use it directly as a parameter as is done in the other create-function for trikom?

apps/payment/models.py
```diff
@@ -362,10 +375,12 @@ def _handle_status_change(self):
             """ Handle the completed payment. Remove delays, suspensions and marks """
             self.payment.handle_payment(self.user)
             self.status = status.DONE
+            self._create_fiken_sale(self.payment_price.price)
```
sklirg: Should this be done before the status is set? Just so that in case something here goes wrong, it's not marked as DONE?

apps/payment/models.py
```diff
@@ -362,10 +375,12 @@ def _handle_status_change(self):
...
         elif self.status == status.REFUNDED:
             self.refunded = True
             self.payment.handle_refund(self)
             self.status = status.REMOVED
+            self._create_fiken_sale(-self.payment_price.price)
```
sklirg: Should this be done before the status is set? Just so that in case something here goes wrong, it's not marked as REMOVED?

apps/payment/models.py
```diff
@@ -553,3 +580,106 @@ def _send_receipt(self, payment_date, subject, description, payment_id, items, t
...
     class Meta:
         default_permissions = ('add', 'change', 'delete')
+
+
+class FikenSale(models.Model):
+    stripe_key = models.CharField(
+        'Stripe key',
+        max_length=20,
+        choices=StripeKey.ALL_CHOICES,
+        default=StripeKey.ARRKOM,
```
sklirg: Is it necessary to have a default key? Should it not always be selected? Minor nitpick I guess :p

apps/payment/models.py
```diff
@@ -553,3 +580,106 @@ def _send_receipt(self, payment_date, subject, description, payment_id, items, t
...
+    )
+    original_amount = models.FloatField()
+    kind = models.CharField(
+        'Fiken Sale kind',
+        max_length=20,
+        choices=FikenSaleKind.ALL_CHOICES,
+        default=FikenSaleKind.CASH_SALE,
+    )
+    paid = models.BooleanField(default=True)
+    created_date = models.DateTimeField(default=timezone.now)
```
sklirg: I'd suggest using `auto_now_add`. That's a Django internal feature instead of setting the value with a function.

It's also marked as non-editable, so people can't just go in and edit it.

https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.DateField.auto_now_add

apps/payment/models.py
```diff
@@ -553,3 +580,106 @@ def _send_receipt(self, payment_date, subject, description, payment_id, items, t
...
+    description = models.CharField(max_length=200)
```
sklirg: Could this be a TextField? Or is there a reason for the 200chars max length attribute?

apps/payment/models.py
oleast: I chose it as a default because other places used Arrkom as the default. I can agree on that it can be better to remove it :)

apps/payment/models.py
oleast: I considered it, and I can't really remember why, but I remember reading arguments against it.
I think I'll change it to `auto_now_add`

apps/payment/models.py
oleast: The 200 char limit comes from the Fiken API. I thought it would be nice to have the limit on the model as well

apps/payment/tasks.py
oleast: That sounds reasonable :smile:

apps/payment/tasks.py
oleast: :+1:

apps/payment/constants.py
oleast: Sure, I just copied this from the original location in the Payment model :sweat_smile:

apps/payment/constants.py
oleast: That sounds reasonable

apps/payment/migrations/0031_fikensale_description.py
oleast: Did not really see that it was added

apps/payment/signals.py
oleast: Soo, maybe. The way I thought about it is that nothing should be changed after it has been created. If I understand it correctly a change should be first removed and created again if something like this happens

apps/payment/models.py
oleast: Because it is not supposed to, it is supposed to use `self.transaction_type` :cold_sweat:

apps/payment/models.py
oleast: Yeah, that sounds reasonable

apps/payment/models.py
oleast: Yeah, that sounds reasonable

oleast: But would it not count anyways, since the save method won't be saved if an error occurs?

apps/payment/signals.py
sklirg: I'd just deny it on the model level for now (like, make the whole model non-editable, at least for non-superusers), and get a better understanding of how this is supposed to work in the long run. Should it be deleted and re-created? Should it be updated in-place? And fix it at a later time :)

apps/payment/migrations/0031_fikensale_description.py
sklirg: :'D

apps/payment/models.py
sklirg: Ok, in that case, it's also present in the other function for trikom.

apps/payment/models.py
sklirg: I'm not sure, I didn't look that deep into the existing payment code, but that might be the case.

apps/payment/models.py
sklirg: I agree.

apps/payment/models.py
oleast: I switched them anyways :man_shrugging:

apps/payment/models.py
oleast: It is supposed to be there, since those will always be for Trikom.

apps/payment/signals.py
oleast: After having a chat with the correct people; it should not be possible to edit. I can't find any other way to do that than marking all fields as `editable=False`?

apps/payment/models.py
oleast: Sure, I can store it as an integer. The problem is that I still have to calculate the stripe fee and vat which will lead to floating point numbers?

apps/payment/models.py
```diff
@@ -367,6 +382,39 @@ def is_refundable_reason(self) -> str:
         can_refund, reason = self.payment.check_refund(self)
         return reason
 
+    def create_fiken_sale(self, amount: int, sale_status: str):
+        ore_amount = amount * 100  # Payment price is in Kr, sale amount is in Øre
+        sale = FikenSale.objects.create(
+            stripe_key=self.payment.stripe_key,
+            original_amount=amount * 100,  # Payment price is in Kr, sale amount is in Øre
```
torjusti: Use `ore_amount`?

PR #2336, opened 2019-09-24
Refactor payment status changes to use pre-save signal

oleast:

- QA / Code Review

## Code Checklist

- [ ] I have added tests for the code I added (tests already cover this)

## Description of changes

Use pre-save signals to handle changes in payment status.
Makes things a bit less bloated in models and handles side effects a bit more clearly?

### DIFF
--- a/apps/payment/appconfig.py
+++ b/apps/payment/appconfig.py
@@ -7,3 +7,6 @@ class PaymentConfig(AppConfig):
 
     def ready(self):
         super(PaymentConfig, self).ready()
+        # The following stops pycharm from nagging about unused import statement
+        # noinspection PyUnresolvedReferences
+        import apps.payment.signals  # noqa: F401
--- a/apps/payment/signals.py
+++ b/apps/payment/signals.py
@@ -1,0 +1,34 @@
+from django.db.models.signals import pre_save
+from django.dispatch import receiver
+
+from . import status
+from .models import PaymentRelation, PaymentTransaction
+
+
+@receiver(signal=pre_save, sender=PaymentRelation)
+def handle_payment_relation_status_change(sender, instance: PaymentRelation, **kwargs):
+    if instance.status == status.SUCCEEDED:
+        """ Handle the completed payment. Remove delays, suspensions and marks """
+        instance.payment.handle_payment(instance.user)
+        instance.status = status.DONE
+    elif instance.status == status.REFUNDED:
+        instance.refunded = True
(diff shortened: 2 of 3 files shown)


apps/payment/signals.py
```diff
@@ -0,0 +1,34 @@
...
+        instance.user.change_saldo(instance.amount)
+
+        """ Pass the transaction to the next step, which is DONE """
+        instance.status = status.DONE
+
+    """ Handle when a transaction is being refunded by Stripe """
+    if instance.status == status.REFUNDED:
+        instance.user.change_saldo(-instance.amount)
+
+        """ Pass transaction to the next strip, which is REMOVED """
```
torjusti: Seems stylistically a little weird to use docstrings instead of comments here?

apps/payment/signals.py
```diff
@@ -0,0 +1,34 @@
...
+        instance.status = status.DONE
+    elif instance.status == status.REFUNDED:
+        instance.refunded = True
+        instance.payment.handle_refund(instance)
+        instance.status = status.REMOVED
+
+
+@receiver(signal=pre_save, sender=PaymentTransaction)
+def handle_payment_transaction_status_change(sender, instance: PaymentTransaction, **kwargs):
+    # When a payment succeeds, ot should be stored to the DB """
```
torjusti: You missed the closing quotations :stuck_out_tongue:

oleast: :upside_down_face:

apps/payment/appconfig.py
```diff
@@ -7,3 +7,6 @@ class PaymentConfig(AppConfig):
 
     def ready(self):
         super(PaymentConfig, self).ready()
+        # The following stops pycharm from nagging about unused import statement
+        # noinspection PyUnresolvedReferences
+        import apps.payment.signals  # noqa: F401
```
torjusti: OK, so I have a more substantial comment this time.

From reading the docs, it seems like this `ready()` method can end up being called multiple times. When testing locally it seems to actually be the case as well, and it seems to cause the signal receivers to actually be registered twice as well. The receipt also seemingly gets sent twice.

Do you know anything about this behavior? Can we make this import only run once, somehow?

oleast: I have no idea about it. I just did it as all the other apps did it

torjusti: Sorry, I can't read. The import might run twice but the listeners only run once.

PR #2337, opened 2019-09-25
Added a message to tell users that adblocks may block events

MrHanSan:

- QA / Code Review

## Code Checklist

## Description of changes
Updated the text displayed if no events are loaded. This is done because some adblocks are blocking the API requests, but we have some trouble reproducing this bug.


henrikhorluck:
I also think this was only a problem for OWF, so this PR might not even be necessary

MrHanSan:
The problem was with the api/vi/events, and this request is done on the front page, so if the problem is blocking of the api then this should be updated. If the problem also exists in OVF then we should look at a solution for that as well.

I am unsure about using "adblocker" or "adblock" as "adblocker" is not propper norwegian, so I think we should just use the english one. Or we can use "Reklame blokkerings applikasjon"

PR #2352, opened 2019-10-07
Add Vimeo ID validator for articles

### LINKED ISSUE #326: Admin/toolkit should verify vimeo IDs
When articles are created there is no check to see if the Vimeo ID actually returns a proper video. A check should be implemented so that users get notified if there is a problem with the given ID.

oleast:

- QA / Code Review

## Code Checklist

- Needs API token in prod env

## Description of changes

Add validator to check if a given Vimeo ID corresponds to an actual Vimeo video.

Resolves #326


apps/article/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+from django.conf import settings
+
+VIMEO_API_TOKEN = settings.VIMEO_API_TOKEN
+VIMEO_API_URL_BASE = 'https://api.vimeo.com'
+VIMEO_API_VIDEOS_URL = f'{VIMEO_API_URL_BASE}/videos'
+
+logger = logging.getLogger(__name__)
+
+
+def check_if_vimeo_video_exists(vimeo_id: str) -> bool:
```
henrikhorluck: Thank you for adding typeings!

oleast:
> Looks good, but do note we currently have videos that are no longer available, should that be something we automatically handle?

There could be a feature for a mommy task that checks that and sends a mails about it or something every day?
But That could be a feature-request :P

PR #2359, opened 2019-10-09
Fix table generated when adding and removing users from event

### LINKED ISSUE #2121: Atendee list bug when adding or removing users

- Bug report

It should look normal where every box and number is under its corresponding column. It should just look normal. Even if we remove users or add users through the dashboard, nothing should change.

screenshot from 2018-09-25 21-07-06

If you remove or add a user through the dashboard menu at the attendee list it will look like the screenshot above. We have randomly two boxes, the x for removing users is all the way to the right and the text "none" is occupying the remove column. If you refresh the site it will go back to expected behaviour, its only after deleteing/adding a user

screenshot from 2018-09-25 20-00-23

Remove or add a user to the attendee list manually.

1. go to the attendee list
2. Add a user by writing their name OR remove a user from the list
## Other information

This might be a bug which I didn't catch when I added "year of study" to the attendee list. I'm not sure if this was an issue before, but since it hasn't been brought up I will assume this is a bug from that pull request of mine

tokongs:

- QA / Code Review

## Code Checklist

## Description of changes
This is makes the table that is generated when adding and removing users from an attendance event look okay.

Resolves #2121


tokongs:
Auto format is nice...

PR #2376, opened 2019-10-18
Unsubscribe mail and resubscribe email when primary email is changed

### LINKED ISSUE #2068: Changing primary email address should change email address used for list subscriptions

- Bug report

If I change my primary email address I expect the mailing list services to be updated with my new address, whilst removing my old address.

The old address is not removed and the new address is not added.

* (Add a new email to an account) (_Only required if there is no alternative email added to the account already_)
* Set the new email address as the primary
* _Notice that the action happens immediately_
    * (Vs. unsubscribing / subscribing to a list, which queries the mailing list service to add/remove addresses.)
* Verify that the old primary address is still on the list
* Verify that the new primary address is absent from the list

## Other information

The settings page tells the user that the primary address is the one used for mailing lists:

> **Adressen som er markert som primær brukes til alle våre e-postlister**, gjenoppretning av passord og andre varsler som sendes fra oss.

The [`set_primary`](https://github.com/dotkom/onlineweb4/blob/v4.24.28/apps/profiles/views.py#L326) function does not handle subscribing/unsubscribing to a list. (Neither does the [`Email model`](https://github.com/dotkom/onlineweb4/blob/v4.24.28/apps/authentication/models.py#L278) nor any signals attached to it.)

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

Fixes #2068

### DIFF
--- a/apps/authentication/signals.py
+++ b/apps/authentication/signals.py
@@ -111,8 +111,8 @@ def remove_online_group_members_from_django_group(
 @receiver(pre_save, sender=Email)
 def re_subscribe_primary_email_to_lists(sender, instance: Email, **kwargs):
     user: User = instance.user
-    jobmail = MAILING_LIST_USER_FIELDS_TO_LIST_NAME.get('jobmail')
-    infomail = MAILING_LIST_USER_FIELDS_TO_LIST_NAME.get('infomail')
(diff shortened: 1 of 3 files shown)


sklirg:
Any thoughts for a one-off script/job to unsubscribe all "actually unsubscribed" people in OW4 to reconsolidate whoever have actually tried to unsubscribe but haven't been unsubscribed because they changed their address.

PR #2333, opened 2019-09-18
Adds website field for career opportunities

VigdisIS:
## What kind of a pull request is this?
Fixes #2292 

## Code Checklist

- [x] The code follows dotkom code style 
- [x] The code passes the defined tests
- [ ] I have added tests for the code I added
- [ ] I have provided documentation for the code I added
- [x] The code is ready to be merged

## (Possible) Breaking Changes

- [x] The changes are backwards compatible

- [ ] I have updated the build configuration
- [ ] These changes requires changes to configuration in production 
    - [ ] I have applied the required changes in production
    - [ ] I cannot apply the required changes in production before this is deployed.

## Description of changes
Adds a website field to the model and makes sure it's displayed properly throughout the website.

## Screenshot

Skjermbilde 2019-09-25 kl  20 20 39

PR #2383, opened 2019-10-31
Career application button

### LINKED ISSUE #2292: Add button to jobb application as a dedicated field to be a part of "Nøkkelinformasjon"
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I am making a new "Karrieremulighet" and I have to insert an external link to the application on the company's website in the application text.

**Describe the solution you'd like**
I want a dedicated field in the dashboard where I can add a link to an external site, and it is shown a part of "Nøkkelinformasjon". 

**Describe alternatives you've considered**
Anything that makes it simple to add an external link and it is clearly visible to users of the site.

**Additional context**

aslakhol:
Closes #2292 

## Code Checklist

## Description of changes
This changes the website element from #2333 to an application link, and displays it in a separate button. 

image


aslakhol:
@mariusaarsnes Does this better solve your need?

FluidSense:
@aslakhol "Søk her" is displayed, even if no application URL has been added. Is this intentional? If so, what is the thought behind it?

aslakhol:
@FluidSense Good catch, I was totally planning on not displaying the button if there was no link, but forgot. I'll take care of it.

apps/careeropportunity/migrations/0014_auto_20191031_1239.py
```diff
@@ -0,0 +1,22 @@
...
+
+    operations = [
+        migrations.RemoveField(
+            model_name='careeropportunity',
+            name='website',
+        ),
+        migrations.AddField(
+            model_name='careeropportunity',
+            name='applicationLink',
+            field=models.URLField(blank=True, null=True, verbose_name='søknadslenke'),
```
henrikhorluck: This migration I am quite sure you need to change, because as far as I know,  we still want to keep the current data.
Perhaps [migrations.RenameField](https://docs.djangoproject.com/en/2.2/ref/migration-operations/#renamefield) does what we want

aslakhol: I have looked into it and I disagree:

1. The field "website" is not the same as "application link". They don't mean the same thing, and the information in website would not work as an application link.

2. Since the website field is brand new there is almost no career opportunities that have this set. And in all cases where it is set, the link is to the homepage of the company and would not work as an application link.

aslakhol:
I made the button green, to make it look more like a button. As this page will be out of date when OWF releases I decided to just phone it in and hard set the colors. We won't change our button color before we release OWF. 

image

PR #2384, opened 2019-11-03
Purge most hard coded mentions of 'Komiteer'

### LINKED ISSUE #1566: Use is_staff rather than membership of group to define staff status
I think it'd be cleaner to check if a user is a staff user account (staff is not the same as administrator / superuser) rather than check for the membership of the group "Komiteer".

Since all committee members are synced to this group, we could make sure any persons synced to the "Komiteer" group has the is_staff flag.

oleast:

- QA / Code Review

## Code Checklist

**This forces dashboard access to committee members _only_, which may exclude some people? But they should not have had access earlier, so maybe not a problem?**

## Description of changes

Remove hard coded mentions of the komiteer group and hard coded relations from named groups to event types.

**Moves the burden of permission checking for poster-orders to the permissions themselves instead of just groups. May need some help in discovering if this works as expected.**

Resolves #1566


myth:
:shipit:

apps/posters/dashboard/utils.py
```diff
@@ -25,9 +30,6 @@ def _handle_poster_add(request, form, order_type):
 
     # Let this user have permissions to show this order
     UserObjectPermission.objects.assign_perm("view_poster_order", request.user, poster)
-    GroupObjectPermission.objects.assign_perm(
-        "view_poster_order", Group.objects.get(name="proKom"), poster
-    )
```
FluidSense: Denne funksjonaliteten er savnet av prokom

oleast: Hvordan skal man gå frem med den når man ikke kan refere til gruppen på den måten?

FluidSense: Kan man gjøre noe á la det samme som `get_poster_admins` gjør? Er det en rettighet som kun prokom har?

PR #2390, opened 2019-11-06
Add fixtures for rule bundles

henrikhorluck:
## Description of changes

Add fixture for grade rules, field of study rule, and rule bundles

This means that you can automatically load the data (which is a mirror of dev/prod) for local testing, by running
```
python manage.py loaddata apps/events/fixtures/Rules.json
```

This is ~very~ _extremely_ helpful for local testing

Note: AFAIK there is zero sensitive information in these bundles, and I think it should be much easier to locally test OW4 stuff

### DIFF
--- a/apps/events/fixtures/Rules.json
+++ b/apps/events/fixtures/Rules.json
@@ -1,0 +1,1002 @@
+[
+  {
+    "model": "events.rule",
+    "pk": 1,
+    "fields": {
+      "offset": 0
+    }
+  },
+  {
+    "model": "events.rule",
+    "pk": 2,
+    "fields": {
+      "offset": 0
+    }
+  },
+  {
+    "model": "events.rule",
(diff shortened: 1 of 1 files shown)


oleast:
I agree that fixtures can be ~very~ _extremely_ helpful when testing locally, and I think we should have fixture of more than just rule-bundles!

But I don't think it is the right course of action to replicate our entire database of instances for our PROD database as fixtures.
I think it would be wise to have a small amount of pre-defined fixtures for common use-cases that we can test with, and maintain in an easier manner?

oleast:
@henrikhorluck Bump

henrikhorluck:
Would be nice but should (probably) 1. use yaml instead of json and 2. have a lot more, perhaps to replace use of `django_dynamic_fixtures`

PR #2391, opened 2019-11-08
Adds some instructions on how to setup onlineweb4

kharann:
Updates the readme to in include instructions on how to set up onlineweb4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ $ git config --global user.email your.github@email.com
 $ git clone git@github.com:dotkom/onlineweb4.git
 $ cd onlineweb4
 ```
-### Setting up the onlineweb4
+### Setting up onlineweb4
 Download dependencies by running
 ```bash
 make
--- a/README.md
+++ b/README.md
@@ -40,8 +40,6 @@ To fire up the dev environment, you should use [Docker](https://docs.docker.com/
 
 There is a `Makefile` located in the root directory. This can be used to manage the application without direct interaction with `docker`, `docker-compose`, et al.
 
-To view output from onlineweb4, run `make logs`. To view output from a specific service (e.g. django), prepend the `make` command with `OW4_MAKE_TARGET=django`.
-
 If you can't use `make`, you can fire up the dev environment by issuing `docker-compose up -d`.
 
 If the site doesn't load properly the first time you are running the project, you might need to restart Docker once by running `docker-compose restart`.
@@ -100,7 +98,13 @@ make start
 And your local version of Onlineweb4 should be available on [http://localhost:8000](http://localhost:8000)!  
 You can stop it with `make stop`
 
-## CD/CI
+To view output from onlineweb4, run `make logs`. To view output from a specific service (e.g. django), use one of the following:
+```bash
+OW4_MAKE_TARGET=django make logs # for django
+OW4_MAKE_TARGET=webpack make logs # for the frontend
+```
+
+## CI/CD
 
 Pushes made to the develop branch will trigger a redeployment of the application on [dev.online.ntnu.no](https://dev.online.ntnu.no).
 
--- a/README.md
+++ b/README.md
@@ -22,36 +22,12 @@ The API uses [Django REST framework](http://www.django-rest-framework.org/)
 ## Installation - Git and repository setup
 ### Setting up your Git user and cloning the repo
 ```bash
-$ git config --global core.autocrlf false
-$ git config --global user.name "<your github username>"
-$ git config --global user.email your.github@email.com
-$ git clone git@github.com:dotkom/onlineweb4.git
-$ cd onlineweb4
+git config core.autocrlf false
+git config user.name "<your github username>"
+git config user.email your.github@email.com
+git clone git@github.com:dotkom/onlineweb4.git
+cd onlineweb4
 ```
-### Setting up onlineweb4
-Download dependencies by running
-```bash
-make
-```
-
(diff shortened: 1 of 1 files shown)


plusk:
A lot of this information already exists in the README, like introducing you to Docker. If you run `make` without Docker, it won't work. I'm all for updating the README and making it more user friendly, and if making it more "todo list"-like is the way, then by all means!
But it would be great if you look through the README right under what you wrote, and either rewrite or remove or drop your changes or whatevs.

README.md
```diff
@@ -20,14 +20,38 @@ The API uses [Django REST framework](http://www.django-rest-framework.org/)
...
+### Setting up the onlineweb4
+Download dependencies by running
+```bash
+make
+```
+
+Lets apply all the django changes to the database by running
+```bash
+# creates migrations by packaging model changes into individual migration files
+make makemigrations
```
FluidSense: Is `make makemigrations` needed? All migrations are in the repository and will be applied by `make migrations`, won't it?

README.md
```diff
@@ -20,14 +20,38 @@ The API uses [Django REST framework](http://www.django-rest-framework.org/)
...
+# applies the migrations/changes to your database
+make migrate
+```
+
+Last but not least, lets create a superuser, a user with all types of permissions.
+```bash
+# bash into the docker container
+make bash-backend
+
+# createa a super user from inside the container
```
FluidSense: Sneaky "a" 👀 
```suggestion
# create a super user from inside the container
```

README.md
kharann: Might be tree! I'll test if it does and then change it

README.md
```diff
@@ -20,14 +20,38 @@ The API uses [Django REST framework](http://www.django-rest-framework.org/)
...
-
+### Setting up your Git user and cloning the repo
 ```bash
 $ git config --global core.autocrlf false
 $ git config --global user.name "<your github username>"
 $ git config --global user.email your.github@email.com
 $ git clone git@github.com:dotkom/onlineweb4.git
 $ cd onlineweb4
 ```
+### Setting up the onlineweb4
```
plusk:
```suggestion
### Setting up onlineweb4
```

README.md
```diff
@@ -20,14 +20,38 @@ The API uses [Django REST framework](http://www.django-rest-framework.org/)
...
 
 ## Installation - Git and repository setup
-
+### Setting up your Git user and cloning the repo
 ```bash
 $ git config --global core.autocrlf false
 $ git config --global user.name "<your github username>"
 $ git config --global user.email your.github@email.com
 $ git clone git@github.com:dotkom/onlineweb4.git
 $ cd onlineweb4
```
henrikhorluck: As you see here, every bash-command is prepended with `$`, which is quite helpful in ensuring that the following statement is the command to input, so I would like for this notation to continue

FluidSense: On the other hand, starting every line with `$` makes the commands harder to copy paste 📋

FluidSense:
Since you're looking at updating the readme, it should probably contain information about creating other users than super user as well.
i.e. use the layout and follow the link which is located in `make logs`.
Perhaps we could even add a command which greps the latest activation link from the logs?
This does not necessarily needs to be a part of this PR however, just found it relevant when looking at how to make setup easier.

README.md
FluidSense: Also, `--global` can be removed. Should we advise people to set this globally when this is a matter of preference and not necessary for the setup of OW4?

kharann: I agree with FluidSense, it becomes harder to copy commands when using them to set up the repo

plusk: The reason the config has been set globally is to also have the config for other dotkom repos. Also, maybe Git was less used in the curriculum previously, making it less of a conflict.

README.md
```diff
@@ -20,14 +20,38 @@ The API uses [Django REST framework](http://www.django-rest-framework.org/)
...
 ```
+### Setting up the onlineweb4
+Download dependencies by running
+```bash
+make
+```
+
+Lets apply all the django changes to the database by running
```
Nicolaad: Should be Let's as in "Let us" , rather than Lets as in "the cop lets you go". Alternatively change it to "Let us"
```suggestion
Let's apply all the django changes to the database by running
```

README.md
```diff
@@ -20,14 +20,38 @@ The API uses [Django REST framework](http://www.django-rest-framework.org/)
...
+```bash
+# creates migrations by packaging model changes into individual migration files
+make makemigrations
+
+# applies the migrations/changes to your database
+make migrate
+```
+
+Last but not least, lets create a superuser, a user with all types of permissions.
```
Nicolaad: Again, should be let's (let us) instead of lets

```suggestion
Last but not least, let's create a superuser, a user with all types of permissions.
```

FluidSense:
Another thing that would be great to have in the readme for beginners is how they view their local OW4-instance! e.g. default url and port

README.md
```diff
@@ -20,14 +20,38 @@ The API uses [Django REST framework](http://www.django-rest-framework.org/)
...
+
+Last but not least, let's create a superuser, a user with all types of permissions.
+```bash
+# bash into the docker container
+make bash-backend
+
+# createa a super user from inside the container
+python manage.py createsuperuser
+```
+Now follow the instructions by typing a name, password and an email.
```
FluidSense: There should be a part saying you have to type `exit` to get back to your terminal, just to make it as clear as possible.

PR #2417, opened 2020-01-17
Revert "Add API for feedback"

oleast:
Reverts dotkom/onlineweb4#2407


oleast:
There proved to be complications with migrating the uuid field

PR #2443, opened 2020-03-04
Add github actions to lint and test

tokongs:

- QA / Code Review

## Code Checklist

## Description of changes
Add a github actions workflow to test and lint the project. Don't know if it follows best practices, but it seems to work.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-Name: Test and build
+name: Test and build
 
 on: [push]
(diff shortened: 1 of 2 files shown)


henrikhorluck:
It appears you should be using a branch from this repo, instead of a fork, so that the actions will run?

tokongs:
Yes, that might be right. I thought only workflows in the default branch would run, but that seems to be an old issue and no longer the case.

PR #2444, opened 2020-03-05
GitHub actions

tokongs:

- QA / Code Review

## Code Checklist

## Description of changes
* Add a github actions pipline to lint and test.
* Add a production ready(i hope) Dockerfile
* Add a pipline to deploy the docker image


henrikhorluck:
Any reason you are not doing lints in parallell with backend-testing? Would love to have MAXIMUM SPEED, even though it already is a _major_ improvement to our current CI/CD

henrikhorluck:
also, looks like there is missing a `Error: Missing repository upload token` for Covecov

tokongs:
Updated it to include the codecov token, and run lint and test in parallel. I guess we'll see in a minute if it is any faster.

Dockerfile.prod
```diff
@@ -0,0 +1,38 @@
+FROM node:latest as webpack
```
henrikhorluck: We should perhaps not run latest node on prod? I think a LTS-version would be more suitable, e.g. `node:12`, would probably be fine either way, just a lil bit worried in terms of compability with our somewhat outdated frontend-dependencies.

Otherwise great stuff! 🍨 

Also: there are several files here that are lacking newlines at the end, and I think we should add it _just_ to be sure, you know?

tokongs: The stage using node is just a build stage, but yes, I agreee. I have updated the PR now.

PR #2445, opened 2020-03-06
Add Mailingslist models

henrikhorluck:

- QA / Code Review

## Code Checklist

- [x] The code is ready to be merged  (Gsuite-sync coming later)

## Description of changes

Add a model so that we don't have to manually change the mail-list-models
The plan is to make a view of this in OWF, through the API, and then remove the page in OW4
### Todo
- [ ] Make view-site in OWF
- [ ] Make the lists sync with Gsuite
- [ ] Remove the `https://online.ntnu.no/resourcecenter/mailinglists/`


apps/mailinglists/models.py
```diff
@@ -0,0 +1,12 @@
...
+from django.utils.translation import ugettext as _
+
+
+class Mailinglist(models.Model):
+    email = models.EmailField(_("epostadresse"), unique=True)
+    name = models.CharField(max_length=100)
+    description = models.TextField(blank=True, null=True)
+    public = models.BooleanField(default=False)
+    contained_emails = models.ManyToManyField(
+        "self", blank=True, related_name="mailinglists_mail_is_in"
```
henrikhorluck: Would love some input on what to name these fields, I am not convinced the current ones are clear enough to singnify what they are. (`contained_emails` == emails that are on this list, `mailinglists_mail_is_in` == self explainatory)

oleast: I think it could be beneficial to separate the mailing lists and the mails that are in the list to separate models?
Having one model to represent the mailinglist itself with all its settings and such, and have another model to represent mails that are listed in it. Of course a thing can be both, but it simplifies the models a lot, w

apps/mailinglists/views.py
```diff
@@ -1,6 +1,17 @@
...
+from .models import Mailinglist
+from .serializers import MailinglistSerializer
+
+
+class MailinglistViewSet(viewsets.ReadOnlyModelViewSet):
+    permission_classes = (AllowAny,)
+    serializer_class = MailinglistSerializer
+
+    def get_queryset(self):
+        return Mailinglist.objects.filter(public=True)
```
oleast: Could be nice to have functionality for editing this?

Using the `rest_framework.permissions.DjangoModelPermission` class to restrict access (still allows viewing without permission).
Could also allow users with permissions to see non-public lists

oleast:
Could we change the rendered lists to also be based on these models?

henrikhorluck:
> Could we change the rendered lists to also be based on these models?

As you can see in the `Todo`-section of the PR, I was thinking this was a perfect opportunity to move this page to OWF, and create the replacement page there, instead of migrating the OW4-page, and limit the frontend-work here to only set up redirects/"This page has been moved" on OW4.

apps/mailinglists/models.py
henrikhorluck: I tried splitting it up into two models now, would be great to have some feedback about potential solutions

apps/mailinglists/views.py
henrikhorluck: Added now!

apps/mailinglists/signals.py
```diff
@@ -0,0 +1,14 @@
+from django.db.models.signals import post_init
+from django.dispatch import receiver
+
+from .models import Mail, Mailinglist
+
+
+@receiver(signal=post_init, sender=Mailinglist)
```
oleast: Do we really want this to be `post_init`?

Won't that mean we will create a `Mail` instance in our db when a MailingList object in python is created? Before is it saved to the database, which means it is potentially invalid or that it may never be saved?

I would think we could use a `post_save` here, and also make sure it is not created again for each time the MailingList model is instantiated/saved, which it looks like this will.

For that we would have to differentiate on when an instance is created and when it is updated, which we'll get from the `created` kwarg.
When updating the existing one we'll have to make sure we still the the unique one even if some or _all_ of the fields are updated

apps/mailinglists/models.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        verbose_name = _("Organisasjons-e-post")
+        verbose_name_plural = _("Organisasjons-e-poster")
+
+
+class Mailinglist(models.Model):
```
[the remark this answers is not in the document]
henrikhorluck: Renamed it to MailGroup instead, I think that is clearer

apps/mailinglists/models.py
```diff
@@ -0,0 +1,64 @@
+from django.db import models
+from django.utils.translation import ugettext as _
+
+
+class Mail(models.Model):
+    """
+    The email of an organization, or person whose email is on one of Online's mailinglists
+    """
+
+    email = models.EmailField(_("e-postadresse"), unique=True)
```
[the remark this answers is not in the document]
henrikhorluck: Renamed it to `Organization`, I think that is clearer

apps/mailinglists/views.py
```diff
@@ -1,6 +1,29 @@
...
+
+
+class MailinglistViewSet(viewsets.ModelViewSet):
+    """
+    E-postlister som brukes av mange forskjellige organisasjoner rundt om i Trondheim.
+    """
+
+    permission_classes = (DjangoModelPermissionsOrAnonReadOnly,)
+    serializer_class = MailinglistSerializer
+
```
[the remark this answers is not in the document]
henrikhorluck: Added some filters now :)

apps/mailinglists/appconfig.py
```diff
@@ -4,3 +4,14 @@
...
+    def ready(self):
+        super().ready()
+        # noinspection PyUnresolvedReferences
+        import apps.mailinglists.signals  # noqa: F401
+
+        from watson import search as watson
+
+        from apps.mailinglists.models import MailGroup
+
+        watson.register(MailGroup, fields=("public",))
```
oleast, line 17: For actual results we should specify the text fields of the model. This boolean field doesn't really make sense in a text-search context

apps/mailinglists/filters.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class WatsonFilter(filters.CharFilter):
+    def filter(self, queryset, value):
+        if value and value != "":
+            queryset = watson_search.filter(queryset, value)
+        return queryset
+
+
+class MailGroupFilter(filterset.FilterSet):
```
oleast, line 16: Would be nice to also be able to filter on the domain

apps/mailinglists/tests.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def test_authenticated_api_call(self):
+        # FIXME: Make this test pass
+        content_type = ContentType.objects.get_for_model(MailGroup)
+        permission = Permission.objects.get(
+            codename="view_mailgroup", content_type=content_type
+        )
+        self.user.user_permissions.add(permission)
+
+        self.assertTrue(self.user.has_perm("mailinglists.view_mailgroup"))
```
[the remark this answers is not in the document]
henrikhorluck: It was mainly used for myself when trying to give the user the correct permissions

apps/mailinglists/tests.py
```diff
@@ -0,0 +1,160 @@
...
+            [
+                OrderedDict(
+                    [
+                        ("id", 1),
+                        ("email", "gresk@linjeforeninger.no"),
+                        ("name", "Greske linjeforeninger"),
+                        ("description", "Linjeforenigner med greske navn."),
+                        ("public", True),
+                        (
+                            "members",
```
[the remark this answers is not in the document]
henrikhorluck: What do you suggest we do instead? I am rather new to API-testing, so would appreciate feedback on good practices 😄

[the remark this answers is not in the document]
oleast: I think it is OK to just check the status code, and maybe check a specific field in some very specific cases where serialization is very complex or dependent on how you do the request. Like if a field is a specific thing when you are logged in and other of you are not.

Otherwise serializer behavior us very predictable, and the only thing you end up doing micro-managing the test data when you are doing changes :stuck_out_tongue:

apps/mailinglists/views.py
```diff
@@ -1,6 +1,44 @@
...
+    queryset = MailGroup.objects.all()
+
+    def get_queryset(self):
+        queryset = super().get_queryset()
+        if not self.request.user.has_perm("mailinglists.view_mailgroup"):
+            queryset = queryset.filter(public=True)
+        return queryset
+
+
+class MailEnitytViewSet(viewsets.ModelViewSet):
```
FluidSense, line 27: Typo "MailEnityt" 👼

FluidSense (commented):
I might just have overlooked it, but when are the new mailing lists actually created in G Suite?

henrikhorluck:
> I might just have overlooked it, but when are the new mailing lists actually created in G Suite?

That is not a part of this PR, I was thinking to make another one afterwards, so that it isn't too much at once

apps/mailinglists/views.py
henrikhorluck: Good catch!

FluidSense (approved):

PR #2446, opened 2020-03-08
Base saldo on transaction history instead of a dedicated field

oleast:

- QA / Code Review

## Code Checklist

 Maybe not?

**The saldo of 7 users will be affected by this change, as there have been race-conditions that have affected the saldo value to not be changed correctly**

## Description of changes

This changes the way a users saldo is stored.
It was previously a value stored for the user, which was updated whenever a transaction was completed or a purchase was made in Nibble.

This is changed to create a transaction for _all_ things which handle money. The saldo is then a calculated value based on all transactions a user has made.

_this will create about 50000 transactions in total_

This also leads to Nibble purchases sending a receipt for the purchased items.


apps/payment/models.py
```diff
@@ -417,52 +427,81 @@ class Meta:
...
+                raise ValueError(
+                    f"Transaction for a shop purchase is not connected to an OrderLine"
+                )
 
     def get_from_mail(self):
         return settings.EMAIL_TRIKOM
 
     def get_to_mail(self):
-        return self.user.email
+        return self.user.primary_email
```
henrikhorluck, line 501: We should be a tad bit careful with this one, but I have checked it now and it appears that `email` is a string either way

apps/payment/transaction_constants.py
```diff
@@ -0,0 +1,12 @@
+class TransactionSource:
```
henrikhorluck, line 1: https://docs.djangoproject.com/en/3.0/ref/models/fields/#field-choices-enum-types

apps/shop/serializers.py
```diff
@@ -18,6 +18,16 @@ class Meta:
 class OrderLineSerializer(serializers.ModelSerializer):
     orders = OrderSerializer(many=True)
 
+    def validate_orders(self, orders: dict):
+        for order in orders:
+            item = Item.objects.get(pk=order.get("object_id"))
+            if not item.available:
+                raise serializers.ValidationError(
+                    "Enklte av de gitte produktene er ikke tilgjengelig"
```
henrikhorluck, line 26: Why is this error-message in Norwegian, while the one in `OrderLine.clean()` was in English?

apps/shop/tests.py
```diff
@@ -6,34 +9,168 @@
...
-    def test_item_list_empty(self):
-        url = reverse("item-list")
+class ShopItemTestCase(APITestCase):
+    basename = "shop_inventory"
+
+    def get_list_url(self):
+        return reverse(f"{self.basename}-list")
 
-        response = self.client.get(url)
+    def get_detail_url(self, _id):
```
henrikhorluck, line 23: There might be a naming-conflict I'm not seeing, or some Python/Django-way of doing things, but why is this marked as unused?

apps/shop/tests.py
```diff
@@ -6,34 +9,168 @@
...
 
-        response = self.client.get(url)
+        response = self.client.get(self.get_detail_url(item.id))
 
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
 
+class ShopOrderLinesTestCase(OAuth2TestCase):
+    scopes = ["shop.readwrite", "read", "write"]
+    basename = "shop_order_lines"
```
henrikhorluck, line 47: Shouldn't this be linked to the actual basename, so that if you wanted to update it you would only have to do it one place?

apps/shop/tests.py
```diff
@@ -6,34 +9,168 @@
...
+    def setUp(self):
+        super().setUp()
+        self.user = G(OnlineUser)
+        G(Email, user=self.user, primary=True)
+        self.access_token = self.generate_access_token(self.user)
+
+    def get_list_url(self):
+        return reverse(f"{self.basename}-list")
+
+    def get_detail_url(self, _id):
```
henrikhorluck, line 58: Same as above regarding `_naming`

apps/shop/tests.py
```diff
@@ -6,34 +9,168 @@
...
+        price = 100
+        quantity = 3
+        item: Item = G(Item, available=True, price=price)
+        self._add_saldo_to_user(price)
+
+        response = self._perform_purchase(
+            [{"object_id": item.id, "quantity": quantity}]
+        )
+
+        self.assertEqual(response.status_code, status.HTTP_406_NOT_ACCEPTABLE)
```
henrikhorluck, line 137: Okay so this was perhaps not added with this PR, but I actually highly disagree that this should be a `406 Not Acceptable`-request. That status code is used for somthing completely [else](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406). If we want to use a status code to check this, I suggest we use `402 Request Failed / Payment Required` in accordance with [Stripe's standards](https://stripe.com/docs/api/errors) , while not _officially_ standard, I think this is an appropiate usecase

apps/shop/tests.py
```diff
@@ -6,34 +9,168 @@
...
+        data = {"user": self.user.id, "orders": order_data}
+        return self.client.post(
+            self.get_list_url(), data=data, **self.generate_headers()
+        )
+
+    def test_insufficient_funds(self):
+        item: Item = G(Item, available=True, price=100)
+        response = self._perform_purchase([{"object_id": item.id, "quantity": 1}])
+
+        self.assertEqual(response.status_code, status.HTTP_406_NOT_ACCEPTABLE)
```
henrikhorluck, line 79: See below

apps/shop/tests.py
```diff
@@ -6,34 +9,168 @@
...
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_saldo_corresponds_with_purchase_amount(self):
+        item: Item = G(Item, available=True, price=100)
+        self._add_saldo_to_user(item.price)
+        self.assertEqual(self.user.saldo, item.price)
+
+        self._perform_purchase([{"object_id": item.id, "quantity": 1}])
+
+        self.assertEqual(self.user.saldo, 0)
```
henrikhorluck, line 105: Should perhaps that the saldo ends up as something other than `0`, all the time? You know just in case it is never adjusted

apps/shop/tests.py
```diff
@@ -6,34 +9,168 @@
...
+
+        self.assertEqual(len(mail.outbox), 0)
+        self._add_saldo_to_user(item.price)
+
+        self.assertEqual(len(mail.outbox), 1)
+
+        response = self._perform_purchase([{"object_id": item.id, "quantity": 1}])
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+        self.assertEqual(len(mail.outbox), 2)
```
henrikhorluck, line 171: I think it would be beneficial to also test that the `to`-address is correct as well, and perhaps that the correct commitee is CC'd?

apps/payment/migrations/0032_auto_20200305_1403.py
```diff
@@ -0,0 +1,25 @@
...
+    for transaction in PaymentTransaction.objects.all():
+        if transaction.used_stripe:
+            transaction.source = "stripe"
+        else:
+            transaction.source = "cash"
+        transaction.save(update_fields=["source"])
+
+
+def remove_data(apps, schema_editor):
+    pass
```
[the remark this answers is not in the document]
oleast: From my experience reverting data migrations will crash if there is no reverse function present?

apps/authentication/models.py
```diff
@@ -242,18 +240,14 @@ def member(self):
...
-        with transaction.atomic():
-            self.refresh_from_db()
-            self.saldo += amount
-
-            if self.saldo < 0:
-                raise NotAcceptable("Insufficient funds")
-
-            self.save()
+    @property
+    def saldo(self) -> int:
```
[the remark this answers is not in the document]
oleast: Could probably add that I guess

apps/payment/models.py
```diff
@@ -417,52 +427,81 @@ class Meta:
         default_permissions = ("add", "change", "delete")
 
 
-class PaymentTransaction(models.Model):
-    """Transaction for a user"""
+class TransactionManager(models.Manager):
+    def aggregate_coins(self, user: User) -> dict:
```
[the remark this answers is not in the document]
oleast: That is probably correct, but it did return a dict at some point :P

apps/payment/models.py
oleast: The `primary_email` field was specifically created to avoid these problems, so that we wouldn't have to worry about that :smile:

apps/payment/models.py
```diff
@@ -417,52 +427,81 @@ class Meta:
...
         return self.datetime
 
     def get_subject(self):
-        return "[Kvittering] Saldoinnskudd på online.ntnu.no"
+        return f"[Kvittering] {self.get_description()}"
 
     def get_description(self):
-        return "saldoinnskudd på online.ntnu.no"
+        if self.source == TransactionSource.STRIPE:
+            return "Saldoinnskudd på online.ntnu.no"
```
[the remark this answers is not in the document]
oleast: I get where you're coming from with that, but the descriptions are supposed to be a bit different from the direct descriptions of the fields themselves. The descriptions are affected by the context of their use.

The description is used in the subject of an email, the `item.name` value is in the context of a receipt, which will be part of the same email, and is just representative of the 'product' you bought. These descriptions will also be used in Fiken

apps/payment/models.py
```diff
@@ -417,52 +427,81 @@ class Meta:
...
-        return items
+        if self.source == TransactionSource.STRIPE:
+            return [{"name": "Påfyll av saldo", "price": self.amount, "quantity": 1}]
+        elif self.source == TransactionSource.TRANSFER:
+            return [
+                {"name": "Overføring av saldo", "price": self.amount, "quantity": 1}
+            ]
+        elif self.source == TransactionSource.CASH:
+            return [
+                {"name": "Påfyll av kontanter", "price": self.amount, "quantity": 1}
```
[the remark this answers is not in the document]
oleast: See comment above

apps/payment/models.py
```diff
@@ -310,7 +312,27 @@ class Meta:
         default_permissions = ("add", "change", "delete")
 
 
-class PaymentRelation(models.Model):
+class StripeMixin(models.Model):
+    status = models.CharField(
+        max_length=30,
+        null=False,
+        choices=status.PAYMENT_STATUS_CHOICES,
```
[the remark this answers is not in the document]
oleast: I chose not to do so as it is a large refactor by itself, and doesn't really have much to do with this PR. This PR was also created before we migrated to Django 3

apps/shop/tests.py
oleast: That may be, but I chose not to change the API, since that would be a breaking change.

apps/shop/tests.py
oleast: See other comment.

apps/shop/tests.py
oleast: Maybe, but I did not find it necessary.
The email things should not be in the domain of the `shop` app, and I don't think anyone is supposed to be CC-ed for a purchase in the shop

apps/shop/tests.py
oleast: Not seeing it :man_shrugging:

apps/shop/tests.py
oleast: The way Django works wont let you do that in all instances anyways, so I don't really see it as that important.

apps/shop/tests.py
oleast: I mean, it is already tested, but not as explicitly.
The insufficient funds tests do so implicitly.

apps/shop/tests.py
oleast: ?

apps/shop/serializers.py
oleast: The validation errors are potentially user-facing, and are mostly written in Norwegian in this project. I just cleared up the anomaly.

apps/shop/tests.py
oleast: Or we could just both have a look at line 101, since it tests that the saldo actually is something else

apps/payment/transaction_constants.py
oleast: I think we could just have that as part of a greater refactor

oleast:
> Okay, lots of comments here. FANTASTIC WORK! Overall I love it   Most of the feedback is somewhat nitpicky, so just be aware that I think you did a great job :)

Thank you very much :smiley: 

> How much will this affect the 7 users you mentioned? And we are certain we have done enough to prevent any race conditions in the future?

How it worked before:

The race-conditions were a result of concurrent requests trying to change the same value at the same moment in time. E.g.

request 1 reads value 100 for saldo
request 2 reads value 100 for saldo (since request 1 isn't finished yet)
request 1 writes 100 + 200 = 300 for saldo
request 2 writes 100 + 50 = 150 for saldo

The user had 100 kr, and made two requests to increase it, one for 200 kr and one for 50 kr. They should then have 350 kr. Because one of the requests effectively overwrites the other they'll end up with 150 kr or 300 kr.

How it works now:

All transactions are separate from each other, adding a transaction can in no way interfere with another in a way that loses data like it did before.
When the saldo is read it will always get the current read-only value from the db.

Affected users:

I've checked their saldo, and none of them should have a negative amount, after this is merged.
The changes will actually make their saldo correct as if there never was any race-conditions in the system.

apps/shop/tests.py
henrikhorluck: By unused, I mean the `_name` naming, which is usually used if you want arguments that are perhaps unused, I was just wondering why it is named that way?

oleast: The `_name` convention is used for private methods and fields

apps/payment/models.py
```diff
@@ -417,52 +427,81 @@ class Meta:
...
-    """Transaction for a user"""
+class TransactionManager(models.Manager):
+    def aggregate_coins(self, user: User) -> dict:
+        """
+        :return: The aggregated amount of coins in a users wallet.
+        """
+        value = (
+            self.filter(user=user, status=status.DONE)
+            .aggregate(coins=models.Sum("amount"))
+            .get("coins", 0)
```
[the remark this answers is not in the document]
oleast: Is fix

apps/authentication/models.py
```diff
@@ -242,18 +240,14 @@ def member(self):
...
+        value = (
+            self.paymenttransaction_set.filter(status=PaymentStatus.DONE)
+            .aggregate(coins=models.Sum("amount"))
+            .get("coins", 0)
```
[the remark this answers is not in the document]
oleast: Is fix

PR #2448, opened 2020-03-08
Oauth2 to OIDC

FluidSense:

Our OIDC-provider also provides Oauth2-authentication.
This ports the shop over to the oidc-provider so we avoid duplicate authentication modules.

- QA / Code Review

## Code Checklist

This breaks all SSO-Clients which uses Oauth2, the endpoint will have to be updated and new keys generated.

- Nibble until https://github.com/dotkom/Nibble2/pull/75 is merged

- [ ] The changes 
are backwards compatible


apps/online_oidc_provider/claims.py
```diff
@@ -42,3 +42,13 @@ class Onlineweb4ScopeClaims(ScopeClaims):
...
+
+    info_nibble = (
+        _("Nibble"),
+        _(
+            "Informasjon om dine kjøp og saldo på Nibble",
+        ),
+    )
+
+    def scope_nibble(self):
+        return {'test':'this data was returned'}
```
FluidSense: I am highly uncertain whether this is necessary or not, and if it is: what it really does..

apps/online_oidc_provider/claims.py
```diff
@@ -42,3 +42,8 @@ class Onlineweb4ScopeClaims(ScopeClaims):
 
     def scope_onlineweb4(self):
         return Onlineweb4Userinfo(self.user).oidc()
+
+    info_nibble = (_("Nibble"), _("Informasjon om dine kjøp og saldo på Nibble"))
```
henrikhorluck: I don't think we want a translation of `Nibble`, if there is one, since it is our own name for it (that is what `_`-does, which is actually the `django.utils.translation.gettext`

apps/online_oidc_provider/authentication.py
```diff
@@ -20,3 +23,56 @@ def authenticate(self, request):
             raise exceptions.AuthenticationFailed("The oauth2 token has expired")
 
         return oauth2_token.user, None
+
+
+class TokenHasScope(BasePermission):
```
henrikhorluck: Is there a reason we are writing our own version of `TokenHasScope`, instead of just using DRF's built in one? If so that should be clearly stated in a comment here

FluidSense: AFAIK DRF does not have a built-in `TokenHasScope`. We previously used oauth2_provider's `TokenHasScope`, made by Django OAuth Toolkit to provide a support layer for DRF. As we are no longer using oauth2_provider, we cannot use their `TokenHasScope` either.

apps/online_oidc_provider/claims.py
FluidSense: Well spotted! As mentioned earlier, this is heavily copy-pasted so I'm not quite certain of whether this code is at all necessary. Whether it is or not, I'll remove the translation-method.

apps/authentication/migrations/0002_auto_20150304_2211.py
```diff
@@ -7,10 +7,7 @@
```
sklirg: I don't think it's that smart to change migrations.

Removing the dependency on this migration doesn't remove the tables/columns from the database. A separate migration is required to do this. After that migration is done, a new change can be done where the requirement is removed, and migrations can be squashed to remove the dependency.

sklirg: I just saw that you've done it correctly with the sso app migration, good. But I still think this migration shouldn't be changed (and therefore all the sso migrations have to be kept), until it is squashed so that sso is fully removed. It's a bit of a long run, but I don't think it matters that much to keep the migrations in for a while. Maybe @duvholt could chime in on this?

apps/online_oidc_provider/authentication.py
sklirg: I'm also a bit vary of this. Could you do some more research with regards to this? Could we depend on oauth2_provider for that function alone, or is it tightly coupled to its own OAuth2 provider? Do we want to depend on it, or do we want to run our own fork of the code? At least get some tests up if we do :)

apps/authentication/migrations/0002_auto_20150304_2211.py
FluidSense: @sklirg I see! So it would be better to not involve this and perhaps add something like it in it's own PR for example?

apps/online_oidc_provider/authentication.py
FluidSense: Hm, I've been digging through the code for the OAuth2 provider, and the answer is a big _I don't know_. I guess I can continue testing some more, but it seems that if we are to use the TokenHasScope from oauth2_provider, we will need duplicates of settings, to ensure that the oidc_provider and oauth2_provider has the same settings, as TokenHasScope relies on settings for oauth2_provider.

FluidSense: The class is by the way taken from this stale issue at oidc_provider's repo https://github.com/juanifioren/django-oidc-provider/issues/342

Other issues related to this worth a read:
https://github.com/juanifioren/django-oidc-provider/issues/242
https://github.com/juanifioren/django-oidc-provider/issues/78

FluidSense:
I am a bit conflicted about this seeing as https://github.com/juanifioren/django-oidc-provider is relatively stale and not well-documented. On the other hand, we are already using it and this would require less manual work (i.e. issuing new tokens and secrets to all current applications) while satisfying our current needs.

oleast:
> I am a bit conflicted about this seeing as [juanifioren/django-oidc-provider](https://github.com/juanifioren/django-oidc-provider) is relatively stale and not well-documented. On the other hand, we are already using it and this would require less manual work (i.e. issuing new tokens and secrets to all current applications) while satisfying our current needs.

I can agree in being a bit conflicted on it, but as of right now it still works. If we end up needing to change we could export to a completly separate OpenID provider without too much hassle I think? We may need everyone to change their passwords at worst?

@Aborysa had a look into setting up KeyCloak earlier, which might still be an alternative. Is it much work to get working as a bare-minimum?

apps/online_oidc_provider/claims.py
FluidSense: Learned recently that an empty dictionary should be the default value. Else, "test" is added as a claim to /userinfo with the returned value.

PR #2457, opened 2020-03-29
Change dependency management to Poetry

henrikhorluck:

- QA / Code Review

## Code Checklist

## Description of changes
Need to update deploy script

There is a minor issue currently, which is that linting on CI is now slower, due to it installing _all_ dependencies (except `prod` and `docs`), while it previously only installed the required deps from `requirements-linting.txt`, this can be fixed by moving all deps to `tox`, but I'm not quite the fan, since it distributes the dependencies


Dockerfile.testbase
```diff
@@ -2,12 +2,16 @@ FROM python:3.7
...
 
 # Install deps
+COPY poetry.lock pyproject.toml /
+
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
     apt-get remove -y curl && apt-get install -y --no-install-recommends \
     nodejs libjpeg-dev ghostscript && \
     npm install -g less && \
-    npm install -g yarn && \
-    pip install tox
```
henrikhorluck: I don't get why this is installing tox, isn't that done from `.drone.yml` anyway?

pyproject.toml
```diff
@@ -0,0 +1,138 @@
...
+whitelist_externals =
+    npm
+    py.test
+    isort
+    flake8
+    black
+commands =
+    isort: isort -rc -c apps middleware scripts utils
+    flake8: flake8 apps middleware scripts utils
+    black: black --check apps middleware scripts utils onlineweb4
```
henrikhorluck: For some reason, Drone is unable to find these "external commands", but Github Actions can find them just fine?

README.md
```diff
@@ -1,22 +1,22 @@
 # Onlineweb 4
-[![Build Status](https://ci.online.ntnu.no/api/badges/dotkom/onlineweb4/status.svg?branch=develop)](https://ci.online.ntnu.no/dotkom/onlineweb4) [![codecov](https://codecov.io/gh/dotKom/onlineweb4/branch/develop/graph/badge.svg)](https://codecov.io/gh/dotKom/onlineweb4) [![Requirements Status](https://requires.io/github/dotkom/onlineweb4/requirements.svg?branch=develop)](https://requires.io/github/dotkom/onlineweb4/requirements/?branch=develop) [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
+[![Build Status](https://ci.online.ntnu.no/api/badges/dotkom/onlineweb4/status.svg?branch=develop)](https://ci.online.ntnu.no/dotkom/onlineweb4) [![codecov](https://codecov.io/gh/dotKom/onlineweb4/branch/develop/graph/badge.svg)](https://codecov.io/gh/dotKom/onlineweb4) [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
```
henrikhorluck: I actually think we shouldn't remove the Requires.io badge, but rather update it as a part of the deployment process (e.g. when we update dev), through the api, since we can generate a `requirements.txt` from Poetry https://python-poetry.org/docs/cli/#export

pyproject.toml
henrikhorluck: This turned out to be quite the complicated task, as Drone doesn't cache _everything_ between steps, only the workspace, so links in `usr/local/bin`, and the downloaded pip-packages are not kept. I solved this by installing during each step, and setting up so that the pip-cache is kept between steps.

.drone.yml
```diff
@@ -95,14 +113,13 @@ steps:
     when:
       event:
         - push
-    group: testing
     commands:
+      - pip install -rrequirements.txt
```
oleast: missing a space?

.drone.yml
```diff
@@ -12,72 +13,89 @@ steps:
...
-    image: registry.online.ntnu.no/dotkom/onlineweb4-testbase:3.7
+    image: registry.online.ntnu.no/dotkom/onlineweb4-testbase:3.7-poetry
+    environment:
+      PIP_CACHE_DIR: ./.pip-cache
     when:
       event:
         - push
-    group: testing
     commands:
+      - sleep 30 # ensure python checks can run in parallell (they share pip-cache)
```
oleast: this seems very error prone and dangerous

.drone.yml
```diff
@@ -12,72 +13,89 @@ steps:
       mount:
         - ./.tox
         - ./node_modules
+        - ./.pip-cache
       volumes:
         - /tmp/cache:/cache
 
   - name: setup
-    image: registry.online.ntnu.no/dotkom/onlineweb4-testbase:3.7
+    image: registry.online.ntnu.no/dotkom/onlineweb4-testbase:3.7-poetry
```
oleast: should probably just replace the testbase?

.drone.yml
henrikhorluck: Yes. This also didn't work because it slept across steps in Drone. I serialized the Drone build now, as a potential solution to this.

oleast:
Mark the changes in #2459 , hehe

.drone.yml
henrikhorluck: I replaced testbase now, I do not think that should affect existing CI, since tox was always downloaded separate anyway

henrikhorluck:
> I think it looks good, but I have no experience with Poetry, so I don't know the fine grained details.
> 
> Is there anything left to change in the build scripts on nansen?

For the build script we need to make sure everything is installed through Poetry, but I think that's it.

PR #2463, opened 2020-04-18
Add filters to career API

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

Add API filters so that searches can be done against the database instead of just in the front-end.

### DIFF
--- a/apps/careeropportunity/filters.py
+++ b/apps/careeropportunity/filters.py
@@ -1,0 +1,32 @@
+from django_filters import filters, filterset
+from watson import search as watson_search
+
+from .models import CareerOpportunity
+
+
+class CharInFilter(filters.BaseInFilter, filters.CharFilter):
+    pass
+
+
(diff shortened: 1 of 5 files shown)


apps/careeropportunity/filters.py
```diff
@@ -0,0 +1,32 @@
+from django_filters import filters, filterset
+from watson import search as watson_search
+
+from .models import CareerOpportunity
+
+
+class CharInFilter(filters.BaseInFilter, filters.CharFilter):
+    pass
```
FluidSense, line 8: Is this necessary?

oleast: Yes, it is kinda strange but it essentially means: _combine these classes together, without adding anything else_

FluidSense: Aha! I see it now, thanks!

FluidSense:
@oleast Could you make an example of a request to test this with? 😄

PR #2466, opened 2020-04-19
Change raven for sentry-sdk

tokongs:

- QA / Code Review

## Code Checklist

## Description of changes
Change out raven for sentry-sdk as this seems to be what Sentry recommends.

Screenshot_20200511_193927


pyproject.toml
```diff
@@ -64,17 +64,17 @@ django-appconf = "~1.0"
...
 
 psycopg2-binary = { version = "~2.8", optional = true }
 uwsgi = { version = "~2.0", optional = true }
-redis = { version = "~3.3", optional = true }
+redis = "~3.3"
 
 # Docs
 sphinx = { version = "*", optional = true }
 sphinx-rtd-theme = { version = "*", optional = true }
+sentry-sdk = "^0.14.3"
```
henrikhorluck, line 77: Since we are using the integrations, I think it would be a little clearer if we show that here, so that if e.g. `sentry-sdk` has a new version of `celery` it doesn't support, we would become aware of it
They seem to not have an extra for redis 🤔 
```suggestion
sentry-sdk = { version = "^0.14.3", extras = [ "django", "celery" ] }
```

templates/base.html
```diff
@@ -224,8 +223,6 @@
     </section>
 
     {% block js %}
-        <script src="https://cdn.ravenjs.com/3.17.0/raven.min.js" crossorigin="anonymous"></script>
```
henrikhorluck, line 227: I think this removes our Sentry-tracking from all frontend-code? We should perhaps migrate to the [Javascript version](https://docs.sentry.io/platforms/javascript/) instead?

templates/500.html
```diff
@@ -1,39 +1,15 @@
...
-        </div>
-    </section>
+<script src="https://browser.sentry-cdn.com/5.15.4/bundle.min.js" integrity="sha384-Nrg+xiw+qRl3grVrxJtWazjeZmUwoSt0FAVsbthlJ5OMpx0G08bqIq3b/v0hPjhB" crossorigin="anonymous"></script>
+
+{% if sentry_event_id %}
+<section>
+<script>
+  Sentry.init({ dsn: '{{ sentry_dsn }}' });
+  
+  Sentry.showReportDialog({ eventId: '{{ sentry_event_id }}' })
```
henrikhorluck, line 12: Could we get a picture how what this looks like?

pyproject.toml
tokongs: I'm not sure I fully understand the extras option, but I guess this is better than what i originally had. Will take a look at it.

templates/base.html
tokongs: Yes. You are right. I didn't think about the JS part of the application. Will fix.

pyproject.toml
henrikhorluck: It’s like saying “we have this extra feature” but that requires this additional dependency, and that dependency might have a specified version, which can lead to compatibility issues, which is what this could potentially guard us against

templates/base.html
```diff
@@ -224,9 +223,10 @@
...
-        <script>Raven.config('{% sentry_public_dsn "https" %}').install()</script>
-
+        <script
+            src='https://js.sentry-cdn.com/df3629be91944be79c790681d9a57f76.min.js'
+            crossorigin="anonymous">
+        </script>
```
henrikhorluck, line 229: I think we need to activate the SDK outside of the 500-template (and with that, I don't think you would need to initialize it in the template), and I think we should add the integrity hash as well

templates/500.html
```diff
@@ -1,39 +1,15 @@
...
-                        <button class="btn btn-success" type="submit">Send mail</button>
-                    </div>
-                </form>
-            </div>
-        </div>
-    </section>
+<script src="https://browser.sentry-cdn.com/5.15.4/bundle.min.js" integrity="sha384-Nrg+xiw+qRl3grVrxJtWazjeZmUwoSt0FAVsbthlJ5OMpx0G08bqIq3b/v0hPjhB" crossorigin="anonymous"></script>
+
+{% if sentry_event_id %}
+<section>
```
henrikhorluck, line 8: I think you forgot to close this `<section>`? Also I would _love_ if you just added something like `height: 90vh` or some alternative to `<main>` perhaps just for this page, so that the background looks a tiny bit better, but great work so far!

templates/base.html
henrikhorluck: Also, why is this from `js.sentry-cdn.com` instead of `browser.sentry-cdn.com`? Also I don't think we need to download the SDK in the 500, if we download it in base

PR #2467, opened 2020-04-19
Add event id to attendance event api

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

We needed the id, so I added the id

### DIFF
--- a/apps/events/models.py
+++ b/apps/events/models.py
@@ -542,6 +542,11 @@ class AttendanceEvent(models.Model):
     payments = GenericRelation("payment.Payment")
 
     @property
+    def id(self):
+        """ Proxy Event id """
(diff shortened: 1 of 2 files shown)


apps/events/models.py
```diff
@@ -541,6 +541,11 @@ class AttendanceEvent(models.Model):
 
     payments = GenericRelation("payment.Payment")
 
+    @property
+    def id(self):
```
henrikhorluck, line 545:
```suggestion
    def id(self) -> int:
```

oleast: Yeah, but this will be inferred by most editors?

henrikhorluck: Probably? But no reason not to add it

PR #2468, opened 2020-04-20
Refactor events api

### LINKED ISSUE #2388: Refactor Events API
**Is your feature request related to a problem? Please describe.**

Our Events API is nested and quite complex and inefficient. It also duplicates data for some endpoints, and may even leak data that maybe should not be available for all users.

The API is also missing parts that are needed to make the events pages of OWF fully functional.

**Describe the solution you'd like**

I would like our Events API to be designed a bit more in line with our apps and models.
It should also strive to be more efficient for different use cases, and more efficiently allow for filtering both in back-end and front-end.

### Lets start with the URLs:

These are the current URLs:

| Model                       | URL   |
|------------------------|-------|
|Events|`/api/v1/events/`|
|Attendees|`/api/v1/attendees/`|
|AttendanceEvents|`/api/v1/registration/attendance-events/`|
|Attendees for users?|`/api/v1/registration/user-attendees/`|

A more fitting URL model would base each URL on each model in the events app:

| Model                       | URL   |
|------------------------|-------|
|Events|`/api/v1/events/events/`| _(Yeah, I know, but kinda required)_
|Attendees|`/api/v1/events/attendees/`|
|AttendanceEvents|`/api/v1/events/attendance-events/`|
|Rule (and all sub-classes)|`/api/v1/events/rules/`|
|RuleBundles|`/api/v1/events/rule-bundles/`|
|Extra|`/api/v1/events/extras/`|
|Reservations|`/api/v1/events/reservations/`|
|Reservee|`/api/v1/events/reservees/`|
|GroupRestriction|`/api/v1/events/restrictions/`|

### The views

Each viewset is connected to a Model and a URL as given above.
Here it is key to keep each model by itself, instead of nesting models inside each other.

#### EventsViewSet

Keep mostly as-is. The queryset filtering is good and gives the users the correct allowed events.
What should be changed:
Serializer should not nest AttendanceEvent or CompanyEvent, nor anything else than maybe the Image.
The filters should be made more efficient with Django `Q` statements and annotations as much as possible instead of filtering in Python.

To make this view more efficient we can utilize a different serializer for `list` and `retrieve` actions.
The `list` action should be efficient, and only deliver required fields at a minimal cost.
The `retrieve` action would only be used to display a single event at a time, and can, therefore, be a bit more inefficient. This serializer could list things like if the user can attend and other things that must be calculated.

#### AttendeesViewSet

This viewset should combine both the current attendee viewsets, and it should be possible to use query params to filter the output based on what the viewset is used for.

#### AttendeanceEventsViewSet

Should list attendance events with queryset filtering based on that in EventsViewSet.

##### Actions

This viewset could include extra actions for checking attendance related thing to the event.

Could have actions for:
**attend** (POST `/api/v1/events/attendance-events/:id/attend/`):
Attends a user to an event and returns the attendee.

**attendees** (GET `/api/v1/events/attendance-events/:id/attendees/`)
Returns a list of the attending attendees with a bit of extra data.
Creates an anonymous `first_name` and `last_name` if privacy setting for the attendee is set to that.
Can probably also include their number in the queue and such?

#### RuleViewSet and RuleBundleViewSet

Rules should list all event rules, as none of them need to be secret.
A rule can also calculate if a user fulfills the rule on the back-end, so that logic is not needed on the front-end.

RuleBundleViewSet could do the same and also display if the user fulfills the rule.
This would make it easy to work with events on the front-end, as the AttendanceEvent would link to a set of RuleBundles, and the front-end can check if the are fulfilled.

...

**Describe alternatives you've considered**

Fix it.

**Additional context**

Nah.

oleast:

- QA / Code Review

## Code Checklist

They are not, but I left the part that is mostly used.
The other parts shouldn't be used many places, which we should be able to just fix.

## Description of changes

Resolves #2388

Split Event API into more pieces, which will make it easier and faster to handle.
Most of the ideas are stated in the issue, but there are some changes.
The API looks like this:
| URL                       | Model or action   |
|---------------------------|-------------------|
|`/api/v1/event/events/`    | Event |
|`/api/v1/event/attendees/` | Attendee list |
|`/api/v1/event/attendees/<pk>` | Attendee detail |
|`/api/v1/event/attendees/administrate` | Change an attendee from an admin perspective |
|`/api/v1/event/attendees/change` | Change an attendee from the users perspective |
|`/api/v1/event/attendees/register-attendance` | Regme / RFID |
|`/api/v1/event/attendance-events/<pk>/`| AttendanceEvent detail, such as if a user can attend  |
|`/api/v1/event/attendance-events/<pk>/register`| Register for an event |
|`/api/v1/event/attendance-events/<pk>/unregister`| Unregister from an event |
|`/api/v1/event/attendance-events/<pk>/attendee`| Get the attendee for the user for that event |
|`/api/v1/event/attendance-events/<pk>/public-attendees`| List of publicly visible attendee, as well as other anonymous |
|`/api/v1/event/attendance-events/<pk>/extras`| List of extra choices for the event |
|`/api/v1/event/attendance-events/<pk>/payment`| Payment object for the event | 
|`/api/v1/event/extras/`| Extras (public list and management for admins) |
|`/api/v1/event/rule-bundles/`| RuleBundls (public list and management for admins) |
|`/api/v1/event/field-of-study-rules/`| FieldOfStudyRule (public list and management for admins) |
|`/api/v1/event/grade-rules/`| GradeRule (public list and management for admins) |
|`/api/v1/event/user-group-rules/`| UserGroupRule (public list and management for admins) |


apps/events/urls.py
```diff
@@ -48,14 +48,3 @@
...
-router.register(
-    prefix="registration/attendance-events",
-    viewset=views.RegistrationAttendaceEventViewSet,
-    basename="registration_attendance_events",
-)
-router.register(
-    prefix="registration/user-attendees",
-    viewset=views.RegistrationAttendeeViewSet,
-    basename="registration_attendees",
-)
```
oleast, line 61: These were mostly not used at all, maybe just payments in OWF. That can be fixed easily

apps/events/tests/api_tests.py
```diff
@@ -1,426 +1,23 @@
...
 from apps.profiles.models import Privacy
 
-from .utils import (
-    attend_user_to_event,
-    generate_event,
-    generate_payment,
-    generate_user,
-    pay_for_event,
-)
+from .utils import attend_user_to_event, generate_event, generate_user
```
oleast, line 14: Almost none of these tests are removed, hat have just been moved and fixed a bit.

apps/events/models.py
```diff
@@ -82,6 +82,30 @@ def get_queryset(self):
...
+            if not user.is_anonymous
+            else Q()
+        )
+        is_visible_query = Q(visible=True)
+        return (
+            self.get_queryset()
+            .filter(group_restriction_query & is_visible_query | is_attending_query)
+            .distinct()
+        )
+
```
oleast, line 108: This was previously part of the main event queryset, it was moved to the manager since I thought it fit better there. That code has not been changed

apps/events/filters.py
```diff
@@ -31,7 +39,8 @@ class EventDateFilter(django_filters.FilterSet):
     can_change = django_filters.BooleanFilter(method="filter_can_change")
     can_attend = django_filters.BooleanFilter(method="filter_can_attend")
     event_type = BaseNumberInFilter(field_name="event_type", lookup_expr="in")
-    companies = BaseNumberInFilter(field_name="companies", lookup_expr="company__in")
+    companies = BaseNumberInFilter(field_name="companies", lookup_expr="in")
```
oleast, line 42: This change comes from the model change

apps/events/api/tests/view_tests.py
```diff
@@ -0,0 +1,237 @@
...
+    UserGroupRule,
+)
+from apps.events.tests.utils import attend_user_to_event, generate_event, generate_user
+from apps.online_oidc_provider.test import OIDCTestCase
+from apps.profiles.models import Privacy
+
+from .utils import generate_attendee
+
+
+class EventsAPITestCase(OIDCTestCase):
```
oleast, line 23: This TestCase was simply moves, the others in this file are new

apps/events/api/tests/attendee_tests.py
```diff
@@ -0,0 +1,393 @@
...
+
+from apps.authentication.models import OnlineUser
+from apps.events.constants import AttendStatus
+from apps.events.tests.utils import attend_user_to_event, generate_event
+from apps.online_oidc_provider.test import OIDCTestCase
+
+from .utils import generate_attendee, generate_valid_rfid
+
+
+class AttendeeTestCase(OIDCTestCase):
```
oleast, line 14: Maybe the top half of this is new

apps/events/api/tests/attendance_tests.py
```diff
@@ -0,0 +1,386 @@
...
+    pay_for_event,
+)
+from apps.online_oidc_provider.test import OIDCTestCase
+from apps.profiles.models import Privacy
+
+from ...models import Extras
+from .utils import generate_attendee
+
+
+class AttendanceEventTestCase(OIDCTestCase):
```
oleast, line 21: Most of this is moved

apps/events/api/register_attendance_serializer.py
```diff
@@ -0,0 +1,159 @@
...
+from django.db.models import Q
+from rest_framework import serializers, status
+from rest_framework.serializers import ValidationError
+
+from apps.authentication.models import OnlineUser as User
+from apps.events.constants import AttendStatus
+from apps.events.models import Attendee, Event
+
+logger = logging.getLogger(__name__)
+
```
oleast, line 13: This is just so damned complex that i moved it to its own file.
We have functionality to replace it, but regme uses this _hard_ and would have to be completly re-written to make that work

apps/events/api/fields.py
```diff
@@ -0,0 +1,20 @@
+from rest_framework import serializers
+
+
+class SerializerUserMethodField(serializers.SerializerMethodField):
```
oleast, line 4: This was moved

apps/events/api/constants.py
```diff
@@ -0,0 +1,46 @@
+ANONYMOUS_ATTENDEE_NAMES = [
```
oleast, line 1: Some are new, some are old

apps/events/api/register_attendance_serializer.py
henrikhorluck: I suggest we rather make a `serializers`-module, split every serializer into it's own file (or whatever makes most sense) and re-export everything from  `__init__.py`. I think that would be a much cleaner solution.
I also think we should do what I suggested to the models.
I know you have said earlier that that isn't "the django way", but it would be a lot more work should we split this up into different apps, which I don't think is something we want, and I think it would be highly beneficial to split these gigantic files (1000+ lines, like come on) compared to the effort it takes.

oleast: I can agree on splitting some of it up, like the serializers. Not that it can be a bit more modular, like splitting them up by use, such as a file for `RuleBundles` and all the `Rules`.
I don't actually think we should re-export the serializers from `__init__.py`, its OK to just import them from where they are.

I've had another look into splitting up models, and the Django Docs actually write about it: https://docs.djangoproject.com/en/3.0/topics/db/models/#organizing-models-in-a-package
It should be very doable, but I still maintain that they should still be somewhat grouped, not all in their own file.
I also think we have to do it all the time, we have plenty of apps that have few enough models for it not to be a problem at all

apps/events/urls.py
FluidSense: Can you test if OWF use them? Could be nice since OW4 relies on OWF for payments.
If it's too much hassle, there's no events with payments or Nibble to update balance for, so this is a good time to test live 👼

onlineweb4/urls.py
```diff
@@ -137,6 +137,9 @@
     urlpatterns += [
         url(r"^events/", include("apps.events.urls")),
         url(r"^dashboard/events/", include("apps.events.dashboard.urls")),
+        url(
+            r"^events-api/", include("apps.events.api.urls")
+        ),  # url is a dummy, but needed to import the file
```
FluidSense, line 142: What does a dummy url imply? It doesn't go anywhere?

apps/events/api/permissions.py
```diff
@@ -0,0 +1,78 @@
...
+        return attendance_event.is_attendee(user)
+
+    @staticmethod
+    def _has_deadline_passed(attendance_event: AttendanceEvent, user: User) -> bool:
+        """ User can only be unattended before the deadline, or if they are on the wait list. """
+        now = timezone.now()
+        return (
+            now > attendance_event.unattend_deadline
+            and not attendance_event.is_on_waitlist(user)
+        )
```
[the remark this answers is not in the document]
oleast: This permission is only used for the `unregsiter` endpoint, which is what a user would use to register.

Since registering people manually as an admin has a different set of constraints that should be another endpoint with another permission.

So yes, admins should be able to bypass this, but not on this endpoint. When registering on the regular page you are equal to everyone else, even when you are an admin

apps/events/api/permissions.py
```diff
@@ -0,0 +1,78 @@
...
+
+from ..models import AttendanceEvent, Attendee
+
+
+class RegisterPermission(permissions.IsAuthenticated):
+    """
+    Check if a user has permission to register for an attendance event.
+    """
+
+    message = ""
```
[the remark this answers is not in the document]
oleast: The message is what the endpoint returns if the permission check fails. Usually you would have a standard message that is returned from the permission check, but we have dynamic messages, so they are assigned manually

apps/events/api/serializers.py
```diff
@@ -0,0 +1,246 @@
...
+
+class RegisterSerializer(serializers.Serializer):
+    """
+    Serializer used when a user attempts to register for an event
+    """
+
+    recaptcha = RecaptchaField()
+    allow_pictures = serializers.BooleanField(default=False)
+    show_as_attending_event = serializers.BooleanField(default=False)
+    note = serializers.CharField(default="", max_length=100)
```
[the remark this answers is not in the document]
oleast: It is a feature where you can leave a note to the event organizers, like special considerations and such. I haven't looked into the history, but it seems like a feature we have always had, but isn't really used?

apps/events/api/tests/attendance_tests.py
```diff
@@ -0,0 +1,396 @@
...
+    basename = "events_attendance_events"
+
+    def setUp(self):
+        self.committee = G(Group, name="Arrkom")
+        self.user = generate_user(username="_user")
+        self.privacy = G(Privacy, user=self.user)
+        self.token = self.generate_access_token(self.user)
```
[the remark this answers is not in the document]
oleast: It comes from the OIDCTestCase, yes. But so does almost everything else in the TestCase, I think it would be a bit much to include it for all that?

apps/events/api/tests/attendance_tests.py
```diff
@@ -0,0 +1,396 @@
...
+            days=2
+        )
+        self.event.attendance_event.max_capacity = 20
+        self.event.attendance_event.save()
+        self.attendee1 = generate_attendee(self.event, "test1", "4321")
+        self.attendee2 = generate_attendee(self.event, "test2", "1234")
+        self.attendees = [self.attendee1, self.attendee2]
+
+    def get_register_url(self, event_id: int):
+        return self.get_action_url("register", event_id)
```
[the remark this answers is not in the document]
oleast: Also OIDCTestCase

apps/events/api/tests/attendance_tests.py
```diff
@@ -0,0 +1,396 @@
...
+    def get_payment_url(self, event_id: int):
+        return self.get_action_url("payment", event_id)
+
+    def test_anonymous_user_can_get_list(self):
+        response = self.client.get(self.get_list_url(), **self.bare_headers)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_anonymous_user_can_get_detail(self):
+        response = self.client.get(
+            self.get_detail_url(self.event.id), **self.bare_headers
```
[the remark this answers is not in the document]
oleast: Also OIDCTestCase

apps/events/api/tests/attendance_tests.py
```diff
@@ -0,0 +1,396 @@
...
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    @mock_validate_recaptcha()
+    def test_user_cannot_register_for_a_non_attendance_event(self, _):
```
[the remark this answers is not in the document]
oleast: That argument comes from the mock decorator. It can be used to dynamically mock during the method call, so it should be possible to use it, so it has to be passed

onlineweb4/urls.py
oleast: The URL passed here is used for urls registered with django url patterns, but the file we are referencing doesn't have any django URL patterns, we just use the DRF router. So it won't be used for any urls, it is a required parameter

apps/events/serializers.py
```diff
@@ -31,136 +15,6 @@ class Meta:
...
-            raise ValidationError("Du kan ikke melde andre brukere på arrangementer!")
-
-        return user
-
-    def validate(self, data):
-        request = self.context.get("request", None)
-        if not request:
-            raise ValidationError(
-                "internal:Request was not passed as context to serializer"
-            )
```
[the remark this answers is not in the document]
oleast: Yes, they are just moved

apps/events/api/views.py
```diff
@@ -0,0 +1,295 @@
...
+        return Response(
+            {
+                "detail": {
+                    "message": f"{attendee.user} er registrert som deltaker. Velkommen!",
+                    "attend_status": AttendStatus.REGISTER_SUCCESS,
+                    "attendee": attendee.id,
+                }
+            },
+            status=status.HTTP_200_OK,
+        )
```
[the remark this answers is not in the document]
oleast: So yes, anyone can access this endpoint... It was designed as our first DRF endpoint I think? It doesn't really follow any good practice on how to do DRF.

Authentication is handled, just in the serializer, and you can't do anything you don't have permission for. It is just that Regme (the rfid app) expects the responses in a specific way, and I don't really want to get into fixing that app.

apps/events/api/views.py
```diff
@@ -0,0 +1,295 @@
...
+        permission_classes=(permissions.DjangoObjectPermissions,),
+        serializer_class=AttendeeAdministrateSerializer,
+    )
+    def administrate(self, request, pk=None):
+        attendee: Attendee = self.get_object()
+        partial = request.method == "PATCH"
+        serializer = self.get_serializer(attendee, data=request.data, partial=partial)
+        serializer.is_valid(raise_exception=True)
+        serializer.save()
+        return Response(data=serializer.data, status=status.HTTP_200_OK)
```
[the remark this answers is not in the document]
oleast: So you have to be an event admin to be able to call this endpoint at all, that is what the permission does. The serializer for this endpoint lets you edit a specific set of fields that admins can edit.
This endpoint would be called from something like dashboard, where you are in the role of an admin. This means there are specific things you can't do from this endpoint that you are actually allowed to do since it is your user. You would have to call another endpoint, just like any other user to change a consent for example.

apps/events/api/views.py
```diff
@@ -0,0 +1,295 @@
...
+        methods=["POST"],
+        permission_classes=(permissions.IsAuthenticated, RegisterPermission),
+        serializer_class=RegisterSerializer,
+    )
+    def register(self, request, pk=None):
+        user = request.user
+        attendance_event: AttendanceEvent = self.get_object()
+        # Check if the recaptcha and other request data is valid
+        register_serializer = self.get_serializer(data=request.data)
+        register_serializer.is_valid(raise_exception=True)
```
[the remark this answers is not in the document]
oleast: The serializer has a `RecaptchaField` which actually verifies the recaptcha with the recaptcha API during validation, yes

apps/events/api/views.py
```diff
@@ -0,0 +1,295 @@
...
+        "event_end",
+        "id",
+        "is_today",
+        "registration_filtered",
+    )
+    ordering = ("-is_today", "registration_filtered", "id")
+
+    def get_queryset(self):
+        user = self.request.user
+        return Event.by_registration.get_queryset_for_user(user)
```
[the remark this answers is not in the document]
oleast: It gets the queryset of events that a user actually has access to see, which is a complex query. I just moved the logic to a better place than the view. Such logic usually is in a models Manager

apps/events/api/views.py
```diff
@@ -0,0 +1,295 @@
...
+
+    @staticmethod
+    def _get_allowed_attendees(user):
+        """
+        A user is allowed to see attendees for their own user, and for events they are organizing.
+        """
+        attendees = get_objects_for_user(
+            user, "events.change_attendee", accept_global_perms=False
+        )
+        attendees |= Attendee.objects.filter(user=user)
```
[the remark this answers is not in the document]
oleast: It is some neat syntax for logical operations on querysets :smile: 

The `|` works as an `OR` operator for querysets. So basically a way to merge the querysets

apps/events/api/register_attendance_serializer.py
oleast: @FluidSense Correct for this file, but some things have been added, and some things have been changed. I just tried to comment in the places which were mostly old code

apps/events/api/tests/attendance_tests.py
oleast: This was moved to here from another place*

apps/events/urls.py
oleast: OWF uses them for event payments, but I have a PR on fixing things to this API already :smile:

apps/events/api/permissions.py
```diff
@@ -0,0 +1,78 @@
...
+        """ User can only be unattended before the deadline, or if they are on the wait list. """
+        now = timezone.now()
+        return (
+            now > attendance_event.unattend_deadline
+            and not attendance_event.is_on_waitlist(user)
+        )
+
+    @staticmethod
+    def _has_event_started(attendance_event: AttendanceEvent):
+        return attendance_event.event.event_start < timezone.now()
```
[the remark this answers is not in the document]
FluidSense: No, it won't

apps/events/api/tests/attendance_tests.py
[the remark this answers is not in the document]
FluidSense: Alright 👌 I spent more time than I'd like to admit searching for the method declaration, thinking that this test would surely fail on Drone and was visibly confused when it didn't 😅 
I think perhaps using `super()` could be good for others like me who might look at this in the future without changing any functionality, but it's definitely not a requirement!

PR #2469, opened 2020-04-22
Replace CoreAPI with OpenAPI and SwaggerUI

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

Django Rest Framework has moved from CoreAPI to OpenAPI, and we upgraded without actually checking compatibility.
SwaggerUI is also much nicer than the CoreAPI docs pages

Main view, each endpoint can be clicked to display how it works
Screenshot from 2020-04-22 09-01-05

Example of a response object
Screenshot from 2020-04-22 09-02-35

Example of a request body
Screenshot from 2020-04-22 09-02-15

### DIFF
--- a/apps/common/rest_framework/mixins.py
+++ b/apps/common/rest_framework/mixins.py
@@ -5,6 +5,12 @@ from utils.metadata import ActionMeta
 
 
 class DefaultDestroySerializer(serializers.Serializer):
+    """
+    Destroy actions usually don't have a serializer, but for automatic schema generation to work
+    all actions need to have a serializer. This is provided as a default for destroy actions
+    for openapi schemas to be rendered correctly.
+    """
+
     pass
 
 
--- a/apps/common/rest_framework/mixins.py
+++ b/apps/common/rest_framework/mixins.py
@@ -1,7 +1,11 @@
 from django.core.exceptions import ImproperlyConfigured
-from rest_framework import viewsets
+from rest_framework import serializers, viewsets
 
 from utils.metadata import ActionMeta
+
+
+class DefaultDestroySerializer(serializers.Serializer):
+    pass
 
 
 class MultiSerializerMixin(viewsets.GenericViewSet):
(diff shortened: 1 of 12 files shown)


apps/common/rest_framework/mixins.py
```diff
@@ -35,6 +39,8 @@ def get_serializer_class_by_action(self, action: str):
             default_class = self.serializer_classes.get("write", default_class)
             if action == "partial_update":
                 default_class = self.serializer_classes.get("update", default_class)
+        if action == "destroy":
+            default_class = DefaultDestroySerializer
```
FluidSense, line 43: What does this do, especially in relation to Swagger?

apps/payment/views.py
```diff
@@ -132,16 +123,11 @@ class PaymentTransactionViewSet(viewsets.ModelViewSet):
...
-
-    def get_serializer_class(self):
-        if self.action in ["list", "retrieve"]:
-            return PaymentTransactionReadOnlySerializer
-        if self.action == "create":
-            return PaymentTransactionCreateSerializer
-        if self.action in ["update", "partial_update"]:
-            return PaymentTransactionUpdateSerializer
-
-        super().get_serializer_class()
```
FluidSense, line 144: Changing the way the viewsets get serializers only affects Swagger, or is the change based on something else?

apps/common/rest_framework/mixins.py
oleast: What the automatic schema generation does is artificially calling the API endpoints to figure auto how they work. When they call the endpoints with the `destroy` action it tries to get the serializer for it.

The problem is that destroy action doesn't actually use a serializer. But for swagger the function for getting the serializer is used anyways, in case someone decides to modify their `destroy` action to use a serializer.

This usually works when a view has defined a single serializer, since it will just get that serializer. But when we define specific serializers for specific actions we usually don't define a serializer for the destroy action, since it is not used. This just sets a default for the schema view to have something to render. I'll add a comment :smile:

apps/payment/views.py
oleast: The functionality is not changed, it was just refactored to a helper method for defining multiple serializers on a viewset

apps/common/rest_framework/mixins.py
FluidSense: I see! But won't this then create a lot of "rubbish" documentation for `delete`-methods which is not in use or which does not use the serializer? Or is this filtered away by Swagger?

oleast: I actually don't know. Looking at the pages it seems to not be there, so I think its good?

PR #2471, opened 2020-04-24
Serialize event id for public attendees

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

We needed a way to determine what event a public attendee belongs to

### DIFF
--- a/apps/events/api/serializers.py
+++ b/apps/events/api/serializers.py
@@ -210,7 +210,14 @@ class PublicAttendeeSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Attendee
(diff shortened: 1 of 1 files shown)


oleast:
> Just because I am curious:
> Is it thus possible to query for public attendees without passing which event you want the attendees for?

No, that is currently not possible. We could set that up, but it would require a separate endpoint instead of nesting it for the event. Maybe that is the reasonable way to go anyways?

FluidSense:
I see! So the API-call already included the event id? Can't the attendees be linked to that without serialization then?

oleast:
Yes, we could have manually added it in the front-end when we got the data. But in the long run I just thought it cleaner for the endpoint to serialize all the required data anyways

PR #2478, opened 2020-05-23
Fix mark registration offset and rule offsets not extending each other

### LINKED ISSUE #2231: Prikk and delayed signup doesn’t work as expected

- Bug report

The prikk timer starts after the delayed signup has ended

The prikk-timer and delayed signup timer starts at the same time (or something similar)

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

Depends on #2477 

Resolves #2231 

Kinda hard to believe we've has an error like this for so many years :thinking: 
I think it has affected a large amount of people

### DIFF
--- a/apps/events/models.py
+++ b/apps/events/models.py
@@ -4,6 +4,7 @@ import logging
 from collections import OrderedDict
 from datetime import datetime, timedelta
 from random import choice as random_choice
+from typing import List
 
 from django.conf import settings
 from django.contrib.auth.models import Group
@@ -33,6 +34,8 @@ from apps.feedback.models import FeedbackRelation
 from apps.gallery.models import ResponsiveImage
 from apps.marks.models import get_expiration_date
 from apps.payment import status as payment_status
+
+logger = logging.getLogger(__name__)
 
 User = settings.AUTH_USER_MODEL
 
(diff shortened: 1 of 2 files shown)


apps/events/tests/model_tests.py
```diff
@@ -304,6 +306,53 @@ def test_user_group_rule(self):
...
+        self.assertEqual(422, response["status_code"])
+
+        # Simulate a day passing for the offset to not have an effect anymore
+        self.attendance_event.registration_start -= timezone.timedelta(hours=24)
+        self.attendance_event.save()
+
+        # User should be able to register for event since offset has passed
+        response = self.attendance_event.is_eligible_for_signup(self.user)
+        self.assertTrue(response["status"])
+        self.assertEqual(212, response["status_code"])
```
henrikhorluck, line 333: I would prefer if the codes are linked to the relevant class-values, since that would make it clearer what kind of value we are looking after here

oleast: I agree that it could be nice, and I almost went for implementing it. Would be a lot more descriptive for users.

I ended up not doing it because it could end up being a lot of work, and not actually needed to correct the problem at hand.

apps/events/models.py
```diff
@@ -477,7 +462,7 @@ class RuleBundle(models.Model):
     user_group_rules = models.ManyToManyField(UserGroupRule, blank=True)
 
     def get_all_rules(self):
-        rules = []
+        rules: List[Union[FieldOfStudyRule, GradeRule, UserGroupRule]] = []
```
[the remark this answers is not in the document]
oleast: By doing this we kinda loose the ability to handle them for their separate specialization, but I guess it doesn't amount for that much in the end.

apps/events/models.py
```diff
@@ -304,65 +307,83 @@ class Rule(models.Model):
...
+            "message": self.not_satisfied_message,
+            "status_code": self.not_satisfied_code,
         }
 
+    def __str__(self):
+        return "Rule"
+
+    class Meta:
+        permissions = (("view_rule", "View Rule"),)
+        default_permissions = ("add", "change", "delete")
```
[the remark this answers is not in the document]
oleast: By having it as an abstract model we loose all ability to handle Rules as their their own queryset. While it may not be needed now, its a not straight forward to get that functionality back, so I don't quite see the reason to remove it not.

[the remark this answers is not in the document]
henrikhorluck: I do not see any reason for us to query over `Rules`, except for some usage-stats?
Could we at least add a comment either here, explaining that this isn't abstract to allow that specific usecase?

apps/events/models.py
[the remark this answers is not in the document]
henrikhorluck: You would either way have to `isinstance(rules[i], FieldOfStudyRule)` to handle their specializations, this just makes you not have to change this as well when adding more rule-types

apps/events/tests/model_tests.py
henrikhorluck: I would love if you took the time to add it! We think we should prioritize making it as clear as possible to new developers (which often includes ourselves)

oleast: I have tried, but I can say by half way that it is a nightmare :upside_down_face:

henrikhorluck: But in this case isn't it just to replace it with`TypeOfRule.some_type_of_code`? Or is there something I'm missing?

oleast: While that works OK it does little for the actual implementation of the rule bundles. You'll get some clarity on the status code meanings themselves, but we end up with a duplication of their actual meanings inside of the _rule satisfaction_ methods anyways. All of those response dictionaries which are passed around and mutated all the time are the things that actually make this confusing and hard to work with. Just adding an Enum with the codes just adds to that without taking away any of the complexity

apps/events/models.py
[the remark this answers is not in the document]
oleast: I just don't think it makes that much of a difference in this case :shrug:

apps/events/models.py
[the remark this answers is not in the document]
oleast: Yeah, you're probably right

apps/events/tests/model_tests.py
henrikhorluck: I think this test is fine, just that I would love to not have to have a tab of the status-codes open at the same time to check that the codes are the correct ones 😅 
I'm not sure what you are talking about about mutating dictionaries? This test just reasignes `response` every time you do a new check?

oleast: You can have a look at #2480 for changes of the signup thingies if you want to

PR #2479, opened 2020-05-23
Fix double emails for webshop payments (and refactor payment related objects)

oleast:

- QA / Code Review

## Code Checklist

## Description of changes

Payments for webshop have sent two receipt emails previously because the OrderLine sent a receipt and the payment relation sent its own receipt.
This makes payments handle related objects/models with an abstract interface to gather needed information about the payment.

This also solves a lot of problems with circular imports which where a result of the _is_type checking

### DIFF
--- a/apps/payment/mixins.py
+++ b/apps/payment/mixins.py
@@ -1,0 +1,83 @@
+from abc import abstractmethod
+from typing import List
+
+from django.conf import settings
+from django.utils import timezone
+
+from apps.authentication.models import OnlineUser as User
+
+
+class PaymentMixin:
+    """
+    Mixin for models related to the a payment.
+    Each abstract method should be overridden to handle payment correctly from that model type.
+
+    To allow payments for new model, just subclass this mixin in the model and implement all methods.
+    """
+
+    @abstractmethod
+    def get_payment_description(self) -> str:
+        return "No description (unsupported content type)"
+
+    @abstractmethod
+    def get_payment_email(self) -> str:
+        """
+        :return the email of the responsible committee/group/user which created or handles the payment.
+        """
+        return settings.DEFAULT_FROM_EMAIL
+
+    @abstractmethod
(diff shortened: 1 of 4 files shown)


apps/payment/mixins.py
```diff
@@ -0,0 +1,83 @@
+from abc import abstractmethod
+from typing import List
+
+from django.conf import settings
+from django.utils import timezone
+
+from apps.authentication.models import OnlineUser as User
+
+
+class PaymentMixin:
```
henrikhorluck, line 10:
```suggestion
class PaymentMixin(ABC):
```
This is an abstract class

apps/payment/mixins.py
```diff
@@ -0,0 +1,83 @@
+from abc import abstractmethod
```
henrikhorluck, line 1:
```suggestion
from abc import ABC, abstractmethod
```
Se below

apps/payment/mixins.py
```diff
@@ -0,0 +1,83 @@
...
+        Handle the effects of payment relation being refunded.
+        """
+        pass
+
+    @abstractmethod
+    def get_payment_receipt_items(self, payment_relation) -> List[dict]:
+        return []
+
+
+class ReceiptMixin:
```
henrikhorluck, line 60:
```suggestion
class ReceiptMixin(ABC):
```

oleast:
> Great! While you are here, it would be great if you added orderings to the models, so that we can slowly, but surely reduce the amounts of warnings we get in CI

I would like to just focus on the issues at hand, and not everything else related to it. That because _way_ too much of a rabbit hole

apps/payment/mixins.py
oleast: I tried this out, and it just created a clusterfuck with the metaclasses in migrations. I'll leave it out for now.

PR #2480, opened 2020-05-23
Refactor signup responses

oleast:

- QA / Code Review

## Code Checklist

Don't think so, but maybe we need more tests?

## Description of changes

Trying to clean up signup logic a bit, don't know if it became any clearer :shrug:

### DIFF
--- a/apps/events/constants.py
+++ b/apps/events/constants.py
@@ -68,7 +68,7 @@ class RegisterStatus(IntegerChoices):
     # Blocked by rule (41x)
     BLOCKED_FOS_RULE = (
         410,
-        "Din studieretning er en annen enn de som har tilgang til dette arrangementet.",
+        "Din studieretning har ikke tilgang til dette arrangementet.",
     )
     BLOCKED_GRADE_RULE = (
         411,
@@ -76,14 +76,14 @@ class RegisterStatus(IntegerChoices):
     )
     BLOCKED_GROUP_RULE = (
         412,
-        "Dine brukergruppe har ikke tilgang til dette arrangementet.",
+        "Dine brukergrupper har ikke tilgang til dette arrangementet.",
     )
     BLOCKED_GENERIC_RULE = 413, "Du har ikke tilgang til å melde deg på arrangementet."
 
     # Blocked by postponed registration (42x)
     POSTPONED_FOS_RULE = 420, "Din studieretning har utsatt påmelding."
     POSTPONED_GRADE_RULE = 421, "Ditt klassetrinn har utsatt påmelding."
-    POSTPONED_GROUP_RULE = 422, "Dine Brukergrupper har utsatt påmelding."
+    POSTPONED_GROUP_RULE = 422, "Dine brukergrupper har utsatt påmelding."
     POSTPONED_GENERIC_RULE = 423, "Du har utsatt påmelding."
 
     # Error status codes
--- a/apps/events/constants.py
+++ b/apps/events/constants.py
@@ -1,3 +1,6 @@
+from django.db.models import IntegerChoices
+
+
 class AttendStatus:
     REGISTER_SUCCESS = "10"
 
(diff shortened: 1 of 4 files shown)


apps/events/constants.py
```diff
@@ -38,3 +41,67 @@ class EventType:
...
+        403,
+        "Du har ikke tilgang til å melde deg på dette arrangementet.",
+    )
+    BLOCKED_IS_ATTENDING = 404, "Du er allerede meldt på dette arrangementet."
+    BLOCKED_SUSPENSION = 405, "Du er suspendert og kan ikke melde deg på."
+
+    # Blocked by rule (41x)
+    BLOCKED_FOS_RULE = (
+        410,
+        "Din studieretning er en annen enn de som har tilgang til dette arrangementet.",
```
henrikhorluck, line 71:
```suggestion
        "Din studieretning har ikke tilgang til dette arrangementet.",
```
To be consistent with the other messages

apps/events/constants.py
```diff
@@ -38,3 +41,67 @@ class EventType:
...
+    )
+    BLOCKED_GRADE_RULE = (
+        411,
+        "Ditt klassetrinn har ikke tilgang til dette arrangementet.",
+    )
+    BLOCKED_GROUP_RULE = (
+        412,
+        "Dine brukergruppe har ikke tilgang til dette arrangementet.",
```
henrikhorluck, line 79:
```suggestion
        "Dine brukergrupper har ikke tilgang til dette arrangementet.",
```

apps/events/constants.py
```diff
@@ -38,3 +41,67 @@ class EventType:
...
+    )
+    BLOCKED_GENERIC_RULE = 413, "Du har ikke tilgang til å melde deg på arrangementet."
+
+    # Blocked by postponed registration (42x)
+    POSTPONED_FOS_RULE = 420, "Din studieretning har utsatt påmelding."
+    POSTPONED_GRADE_RULE = 421, "Ditt klassetrinn har utsatt påmelding."
+    POSTPONED_GROUP_RULE = 422, "Dine Brukergrupper har utsatt påmelding."
```
henrikhorluck, line 86:
```suggestion
    POSTPONED_GROUP_RULE = 422, "Dine brukergrupper har utsatt påmelding."
```

apps/events/models.py
```diff
@@ -472,18 +444,21 @@ def get_all_rules(self):
...
-            response = rule.satisfied(user, registration_start)
-            if response["status"]:
-                return [response]
-            else:
-                errors.append(response)
+            status, offset = rule.satisfied(user, registration_start)
+            # Return early if any of the responses is allowed for register.
+            # We won't have to care about the rest since we know the user can register.
+            if RegisterStatus.is_allowed(status):
+                return [(status, offset)]
```
henrikhorluck, line 457: If the user is allowed to register, is it right to assume that the offset is guaranteed to be zero, in which case we can just write 0 here?

apps/events/models.py
```diff
@@ -885,85 +819,91 @@ def get_minimum_rule_offset_for_user(self, user: User):
...
-            return future_response
-        if smallest_offset > timezone.now() and offset_response:
-            return offset_response
-        if errors:
-            return errors[0]
-
-    def rules_satisfied(self, user):
+                    raise RegisterException(
+                        status=RegisterStatus.BLOCKED_MARK,
+                        offset=postponed_registration_start,
```
henrikhorluck, line 855: As far as I can see, `offset` is supposed to be a `timedelta`, but here `postponed_registration_start` I think is a `datetime`? Since `self.registration_start` is a `datetime`?
Which might have been an uncaught error from the previous change?

henrikhorluck:
Haha, I forgot this PR so ended up doing it again https://github.com/dotkom/onlineweb4/pull/2704

PR #2481, opened 2020-05-24
Upgrade dependencies

henrikhorluck:

- QA / Code Review

## Code Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,50 +10,49 @@ python = "~3.7"
 Django = "~3.0"
 djangorestframework = "~3.11"
 django-filter = "~2.2"
-pillow = "~7.0"
+pillow = "~7.1"
 markdown2 = "~2.3"
-pytz = "^2019.3"
+pytz = "^2020.1"
 python-memcached = "~1.59"
-stripe = "~2.37"
+stripe = "~2.48"
 icalendar = "~4.0"
 google-api-python-client = "~1.8"
 celery = "~4.4"
-PyMuPDF = "~1.16"
-django-filebrowser = { git = "https://github.com/sehmaschine/django-filebrowser.git", rev = "4c1a86d67cd7424df57b0e6880152b309823f009" }
-# ^ For Django 3 support
+PyMuPDF = "~1.17"
+django-filebrowser = "~3.13"
 django-chunks = { git = "https://github.com/dotKom/django-chunks.git", rev = "e6bf109d4fd286964024c140ac1b8f3d86713540" }
 django-crispy-forms = "^1.9"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -10,50 +10,49 @@ python = "~3.7"
...
 django-guardian = { git = "https://github.com/django-guardian/django-guardian.git", rev = "dba092f5db86808713433388717dcfbaaa928a0b" }
 # ^ TODO: Update when django-guardian > 2.2 is released with fix for `get_objects_for_user` ref. dotkom/onlineweb4#2459
-django-taggit = "~1.2"              # Generic multi-model tagging library
+django-taggit = "~1.3"              # Generic multi-model tagging library
 django-taggit-serializer = "~0.1"   # REST Framework serializers for Django-taggit
 APScheduler = "~3.6"                # Scheduler
 redwine = {git = "https://github.com/dotkom/redwine.git", rev = "1.3.2"}
 reportlab = "~3.5"
-pdfdocument = "~3.3"
+pdfdocument = "~4.0"
```
henrikhorluck, line 39: They changed the default font size from 8 to 9, which _might_ change the size of the text is some tables we are creating

pyproject.toml
```diff
@@ -10,50 +10,49 @@ python = "~3.7"
...
+docutils = "~0.16"                  # Allows django-admindocs to generate documentation
+requests-oauthlib = "~1.3"          # Simplify OAuth2 connections to external apps
+oic = "~1.2"			            # OpenID Connect Client
 python-dateutil = "~2.8"            # Simplify date parsing
 coreapi = "~2.3"                    # API Documentation
-requests = { version = "~2.22", extras = [ "security" ] }
+requests = { version = "~2.23", extras = [ "security" ] }
 # Wiki
-wiki = "0.6b1"
+wiki = "0.6b2"
```
oleast, line 55: Have you had a look at the wiki with this version?
Django-wiki upgrades to Bootstrap 4, which might break the styling for parts of it since we use bootstrap 3

pyproject.toml
oleast: I think it will probably be OK

pyproject.toml
henrikhorluck: Oh no, that is quite the breaking change, they even now require `popper.js` for dropdowns :(

oleast: Yeah, but it was very much about time that they did the upgrade.

I don't think we'll ever upgrade OW4 to Bootstrap 4, so I think we can't ever upgrade this until we either remove the main front-end from OW4, or split the wiki into its own service

PR #2485, opened 2020-07-14
Ensure new expiration date isn't in the past when applying for a membership extension

### LINKED ISSUE #1399: Updated membership "fails" for users with membership > 1 year old
Currrently the new expiration date for memberships is determined like this:

``` python
new_expiration_date = datetime.date(membership.expiration_date.year + 1, 9, 16)
```

Source: https://github.com/dotKom/onlineweb4/blob/develop/apps/approval/views.py#L86

If a user has membership that expired in 2014, they will get 2015 as the new date, which is still an expired date.

Jorrre:
Set new expiration date to 16/09 this year/next year (depending on when the request was sent) instead of 16/09 the year after the previous membership ended. Fix #1399 

- Bugfix
- QA / Code Review

## Code Checklist

## Description of changes
Changed the way the new expiration date is set when applying for a membership extension


apps/approval/views.py
```diff
@@ -116,9 +116,14 @@ def create_membership_application(request):
             )
             return redirect("profiles_active", active_tab="membership")
 
-        # Extend length of membership by 1 year
-        membership = AllowedUsername.objects.get(username=request.user.ntnu_username)
-        new_expiration_date = datetime.date(membership.expiration_date.year + 1, 9, 16)
+        # Grant membership until 16th of September this year if the request was sent previous to 1st of July,
+        # or until 16th of September next year if the request was sent after 1st of July
```
henrikhorluck, line 120: Note: due to leap-years and such, we are not actually guaranteed for it to last until 16th of September the next year. I do not think this is something that we need to address

Jorrre: Alright, so should it just specify that the applicant is granted membership for one year?

henrikhorluck: No, I think this is fine!

PR #2495, opened 2020-07-22
Add superlinter

henrikhorluck:

- QA / Code Review

## Code Checklist


henrikhorluck:
It only lints if there are actually changes to those files, and it is always useful to have a helping hand, giving you confidence that the code you are merging is actually (somewhat) correct.
As you can see from https://github.com/dotkom/onlineweb4/runs/899834060, all our current files pass the CI either way (except for a readme in a package, which I gently ignored, since that really isn't that important)

henrikhorluck:
The superlinter also support listing e.g. javascript, typescript, and python files as well, but we already have tooling for that, so I left it out

PR #2496, opened 2020-07-29
Make PR template more fluid

henrikhorluck:
- Several fields are no longer relevant due to CI/CD which would check those boxes automatically
- Description of changes of changes is important, so moved it to the start

- QA / Code Review

## Code Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,17 +1,11 @@
-## What kind of a pull request is this?
-
-- QA / Code Review
-<!-- Add other options if appropriate -->
-
+## Description of changes
+<!-- It is often obvious what changed by looking at the code, so it is more helpful to say _why_ it should be changed -->
+<!-- If the Pull Request is not ready to be merged, please use a draft pull request -->
 
 ## Code Checklist
 
-- [ ] The code follows dotkom code style 
-- [ ] The code passes the defined tests
 - [ ] I have added tests for the code I added
 - [ ] I have provided documentation for the code I added
-- [ ] The code is ready to be merged
-
 
 ## (Possible) Breaking Changes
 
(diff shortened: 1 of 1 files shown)


henrikhorluck:
Uhm stripe-api-response change that breaks tests?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,11 @@
...
+## Description of changes
+<!-- It is often obvious what changed by looking at the code, so it is more helpful to say _why_ it should be changed -->
+<!-- If the Pull Request is not ready to be merged, please use a draft pull request -->
 
 ## Code Checklist
 
-- [ ] The code follows dotkom code style 
-- [ ] The code passes the defined tests
 - [ ] I have added tests for the code I added
 - [ ] I have provided documentation for the code I added
```
aslakhol, line 8: Keep this for now. But we should probably have a talk about documentation again. I can't remember ever writing documentation for a OW change, for good and for bad.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,11 @@
...
-- [ ] The code is ready to be merged
-
 
 ## (Possible) Breaking Changes
```
aslakhol, line 10: Rather than a mess of checkboxes I think we could use just a comment that said something like "If you have made any possibly breaking changes, please describe them and what changes should be made to production" (Just phrased better).

aslakhol: Basically, it feels like shit to have tons of checkboxes that you don't check. I'd rather remove them if most people shouldn't use them.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,17 +1,11 @@
...
-
+## Description of changes
+<!-- It is often obvious what changed by looking at the code, so it is more helpful to say _why_ it should be changed -->
+<!-- If the Pull Request is not ready to be merged, please use a draft pull request -->
 
 ## Code Checklist
 
-- [ ] The code follows dotkom code style 
-- [ ] The code passes the defined tests
 - [ ] I have added tests for the code I added
```
aslakhol, line 7: Ref the comment I make below:
In this case most people should make tests so I agree with having a checkbox here. It makes the mental difficulty of merging code with no tests higher.

.github/PULL_REQUEST_TEMPLATE.md
henrikhorluck: Agree! Changed it to a new comment now

PR #2500, opened 2020-08-09
Fix stripe API test breakage

henrikhorluck:

- QA / Code Review

## Code Checklist

### DIFF
--- a/apps/payment/tests/api_tests.py
+++ b/apps/payment/tests/api_tests.py
@@ -212,7 +212,7 @@ class PaymentRelationTestCase(OIDCTestCase):
 
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertIn(
-            f"No such PaymentMethod: {fake_payment_method_id}", response.json()[0]
+            f"No such PaymentMethod: '{fake_payment_method_id}'", response.json()[0]
         )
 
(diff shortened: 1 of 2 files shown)


henrikhorluck:
2500 PR's whooo 🍰

apps/payment/tests/api_tests.py
```diff
@@ -212,7 +212,7 @@ def test_event_payment_fails_with_fake_payment_method_id(self):
 
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
         self.assertIn(
-            f"No such PaymentMethod: {fake_payment_method_id}", response.json()[0]
+            f"No such PaymentMethod: '{fake_payment_method_id}'", response.json()[0]
```
aslakhol, line 215: Are these actually supposed to be strings, they seem like they should be the variable?

henrikhorluck: They are f-strings right now, with `fake_payment_method_id` as a variable? The problem is that Stripe apparently changed their API, so that the ID is quoted in `''` now, see https://github.com/dotkom/onlineweb4/runs/954000945

aslakhol: Ah, well yes. That totally doesn't make things more confusing stripe.

PR #2502, opened 2020-08-14
Make committee-applications work again

henrikhorluck:
## Code Checklist

- [ ] I have added tests
- [x] I have provided documentation

## Visual changes

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

PR #2503, opened 2020-08-14
Fix CommitteeApplicationPeriod bug with overlaps

kharann:
## Description of changes
When serializing or changing a CommitteePeriod it will check if the period is overlapping with others. The issue here is that it will check all periods if there is any overlap. However since it checks all overlaps the period will always be in conflict with itself. I changed the code to throw error whenever theres 2 or more overlapping issues instead of max 1 overlap. 

I tested this localy and it works.


kharann:
> Burde de ikke være exclude?

Du har helt rett i det. Var stuck en god stund fordi jeg måtte finne en måte å hente ut primary key fra formen

kharann:
fixed by #2502

Issue #2501, opened 2020-08-09
Updating "opptaksperiode" throws error in dashboard 


monikahm:
**Describe the bug**
Trying to save changes on "opptaksperiode" throws "Opptaksperioder kan ikke overlappe med hverandre".

**To Reproduce**
1. Go to 'dashboard -> komitesøknader -> se opptaksperioder'
2. Click on any application period -> edit -> (eventually change commitee) -> save
3. See error

**Expected behavior**
Updating event even if nothing is changed

**Desktop (please complete the following information):**
 - OS: Windows10
 - Browser Chrome
 - Version 84.0.4147.105

**Additional context**
It works to update application period in administration panel

henrikhorluck:
Fixed in #2504

PR #2504, opened 2020-08-14
Exclude itself (CommiteeApplicationPeriod) when checking for overlapping periods

kharann:
Dashboard will have an validationerror everytime you try to update because everytime you update it will check for overlapping periods. Since the Manager checks towards all periods it will always get a conflict with itself. This PR just makes sure to remove themselves from overlapping periods so it doesnt check up against itself

### DIFF
--- a/apps/approval/api/serializers.py
+++ b/apps/approval/api/serializers.py
@@ -42,7 +42,7 @@ class CommitteeApplicationPeriodSerializer(serializers.ModelSerializer):
 
         overlapping_periods = CommitteeApplicationPeriod.objects.filter_overlapping(
             period.start, period.actual_deadline_method()
-        )
+        ).exclude(pk=self.instance.pk)
 
         if overlapping_periods.exists():
             raise serializers.ValidationError(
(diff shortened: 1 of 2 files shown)


kharann:
This should fix #2501

apps/approval/api/serializers.py
```diff
@@ -42,7 +42,7 @@ def validate(self, attrs: dict):
 
         overlapping_periods = CommitteeApplicationPeriod.objects.filter_overlapping(
             period.start, period.actual_deadline_method()
-        )
+        ).exclude(pk=self.instance.pk)
```
henrikhorluck, line 45:
```suggestion
        ).exclude(pk=period.pk)
```

apps/approval/dashboard/forms.py
```diff
@@ -21,7 +21,7 @@ def clean(self):
         actual_deadline = period.deadline + period.deadline_delta
         overlapping_periods = CommitteeApplicationPeriod.objects.filter_overlapping(
             period.start, actual_deadline
-        )
+        ).exclude(pk=self.instance.pk)
```
henrikhorluck, line 24:
```suggestion
        ).exclude(pk=period.pk)
```

kharann:
period doesnt have a pk (it is null) because its just a copy of the form data. :((

apps/approval/api/serializers.py
kharann: wont work since period doesnt have  pk

Issue #2319, opened 2019-09-05
Payments with deadlines will enforce payment before actual deadline


oleast:
**Describe the bug**
Payments with deadlines will enforce the payment penalties at 8 in the morning the day of the deadline, not accounting for the time of day set on the deadline.

Enforcing payments is run as a background task every morning at 08:00. This task checks all event payments to make sure everyone has paid before the deadline.
The deadline check only uses the actual date of the deadline, and not the time: https://github.com/dotkom/onlineweb4/blob/b50be974e03fd47dcc78ac51e4a78f14fb0f9f31/apps/payment/mommy.py#L41

**To Reproduce**
1. Create an event with attendance
2. Create a payment with a deadline at some date with time after 08:00
3. Wait until 08:00 the date of the deadline
4. Receive mark and mail telling you have not paid, even though the deadline has not passed

**Expected behavior**
Payments should not be enforced before the stated deadline. This will confuse users and give them unexpected marks/suspensions.

**Additional context**
This has happened with the 'Bankett itDAGENE 2019' (`id: 737`).
The deadline was (is) on 05.09.2019 at 23:59:59, while the enforced deadline was at 08:00:00 the same day.

PR #2524, opened 2020-09-12
Fix Payment deadline bug where payments will be enforced before actuall deadline

kharann:
## Description of changes
#2319 

TL:DR; Deadlines will be checked 08:00 every morning. If a deadline is 13 september 00:00, it will actually enforce payment 16 hours before the actual deadline. This is because we used to calculate deadlines in days instead of time. 

I basically changed it to calculate with seconds instead.


aslakhol (approved):
LGTM

Issue #2526, opened 2020-09-14
Most notifications are not sent to their recipients


oleast:
Most notifications are left as not sent to their users, which means users may not receive payment reminders and other important information.

Error event in Sentry:
https://sentry.io/organizations/dotkom/issues/1885443442/events/?project=204971

Lift of notifications where you can see what has not been sent:
https://online.ntnu.no/admin/notifications/notification/?p=17

FluidSense:
Mails are now fixed! Push notifications still remain a mystery.

oleast:
For the actual problem this issue referenced, this has been fixed now. Right @FluidSense ?

FluidSense:
Seemingly!

PR #2527, opened 2020-09-16
Set countdown on celery message

FluidSense:
## Description of changes

Wait 5 seconds before sending notification/mail.
A lot of Notifications fail with "Notification.DoesNotExist", but it exists right afterwards in the admin panel. Let's see if this fixes it!

Related to #2526


oleast (approved):
Why not try it, fix the linting and its good 😄

PR #2528, opened 2020-09-16
Retry on ObjectDoesNotExist

FluidSense:
5 Seconds didn't help, let's try to reschedule the notification task on ObjectDoesNotExistException with a minute delay.
Related to #2526

### DIFF
--- a/apps/notifications/tasks.py
+++ b/apps/notifications/tasks.py
@@ -2,6 +2,7 @@ import json
 import logging
 
 from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
 from django.core.mail import send_mail
 from pywebpush import WebPushException, webpush
 from rest_framework import serializers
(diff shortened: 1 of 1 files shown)


oleast (approved):

apps/notifications/tasks.py
```diff
@@ -109,7 +110,11 @@ class Meta:
         )
 
 
-@celery_app.task(bind=True)
+@celery_app.task(
+    bind=True,
+    autoretry_for=(ObjectDoesNotExist,),
```
oleast, line 115: The actual exception we see is `Notification.DoesNotExist`, will this work for that as well?

FluidSense: https://docs.djangoproject.com/en/3.1/ref/exceptions/#objectdoesnotexist
Yeah! It might not need to be this general tho

oleast: If it works it works 🎉

oleast (approved):

PR #2530, opened 2020-09-18
Add endpoint to use djangos logout

FluidSense:
## Description of changes

External systems should be able to log a user out completely and thus require a new login to OW4.
This endpoint uses django's own logout process to remove the session.

## Code Checklist

- [ ] I have added tests


FluidSense:
This already exists! `/openid/end-session/`

PR #2535, opened 2020-09-22
Set is_staff on node commitees as well

FluidSense:
## Description of changes

This is a minor fix to allow node commitee members to get an Online-mail as well.
This should probably be changed so that a flag can be set on a group instead so that other groups can get e-mails.

## Code Checklist

- [ ] I have added tests


FluidSense:
Should probably run a shell script and loop through all users and set `user.is_staff = user.is_committee`

aslakhol (approved):

PR #2536, opened 2020-09-24
WIP: Change our swagger docs page to use DRF_YASG

kharann:


kharann:
too anoying to work with :D

PR #2537, opened 2020-09-25
Make all endpoints visible for the public

kharann:
You still need authentication to use most of them


FluidSense (approved):
Neato burrito 🌯

PR #2541, opened 2020-09-27
Remove CSRF-requirement for SameSite

FluidSense:
When having OWF on the same domain as OW4, allowing session authentication for DRF requires CSRF which prevents regular API-calls from OWF.

## Code Checklist

- [ ] I have added tests


kharann (approved):
Lgtm

PR #2538, opened 2020-09-25
Add OIDC & API to Readme

FluidSense:
OIDC and API information could be nice for easier integration with OW4, so I added it to the readme

PR #2542, opened 2020-09-27
Add id to serializer

Paalar:
## Description of changes
Add ID to profile serializer.
There is no way to get the user's ID. You can get it from `public/profile`, but a user may hide their profile, so it is not a sure fire way to get the ID.

### DIFF
--- a/apps/profiles/serializers.py
+++ b/apps/profiles/serializers.py
@@ -16,6 +16,7 @@ class ProfileSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
         fields = (
+            "id",
             "first_name",
             "last_name",
             "username",


kharann (approved):
Yep yep looks good

FluidSense:
@Paalar If you are authenticated through the OpenID Connect Provider, you automatically get the ID of the user in the response

Paalar:
@FluidSense when using the authorization code flow with pkce there is no id returned.

FluidSense:
> @FluidSense when using the authorization code flow with pkce there is no id returned.

@Paalar Hm, is there no key named "sub" in the return from the `token`-endpoint? I seem to be getting the id at the key "sub", but I am currently checking different flows to see how the data differ!

Paalar:
@FluidSense The response does not contain a `sub` key. This is an example of what is returned:
```
{
"access_token": "",
"refresh_token": "",
"token_type": "bearer",
"expires_in": 3600,
"id_token": ""
}
```

FluidSense:
@Paalar Yes, I see it now! My script had an automated fetch to `/openid/userinfo`,
If you want to get the ID as part of the login process, you can request it from there after acquiring the access token.

That is however unrelated to the changes in this PR, I do agree that there is no point in hiding the ID when retrieving the profile!

Paalar:
This process should definitely be written somewhere including all the scopes.
That could be a solution, but I still think not retrieving the unique identifier of the object is weird, as you agreed with.

FluidSense:
It is defined in the [OIDC-spec](https://openid.net/specs/openid-connect-basic-1_0-28.html#userinfo), but I agree that searching through there is not a good way to develop and find documentation!

That's why I'm working on #2538!

PR #2546, opened 2020-10-07
exclude migrations from black

aslakhol:
Added exclusion settings for black.


aslakhol:
Making draft pending the outcome of a discussion in slack.

henrikhorluck:
Superledere by #2547

PR #2547, opened 2020-10-07
Automatically format migration files

henrikhorluck:
## Description of changes

See discussion on Slack, and #2546 

## Code Checklist

- [ ] I have added tests

## Visual changes

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>


Makefile
```diff
@@ -31,9 +31,18 @@ logs:
...
+# the migration-files that Django output are not formatted according to Black code-style
+# there is no plan to add that as a feature to Django ref. https://code.djangoproject.com/ticket/24275
+# therefore we have modified the default one to also format _only_ the files that Django created
+# which is why we copy the output of the interactive `makemigrations`-command into a temporary file
+# and then search for the lines containing the paths to the migrations, and send them to Black with xargs
+makemigrations-no-format:
 	@docker-compose run --rm $(BACKEND_SERVICE_NAME) python manage.py makemigrations $(OW4_MAKE_TARGET)
 
+makemigrations:
+	@docker-compose run --rm $(BACKEND_SERVICE_NAME) bash -c "python manage.py makemigrations $(OW4_MAKE_TARGET) 2>&1 | tee /tmp/created_migrations.log && grep '^[ \t]*apps/' /tmp/created_migrations.log | xargs black"
```
sklirg, line 44: You can avoid writing to a temporary file and simply grep on the output directly
`python manage.py makemigrations $(OW4_MAKE_TARGET) 2>&1 | grep '^[ \t]*apps/' | xargs black`

henrikhorluck: That doesn't show the output of `./manage.py makemigrations` in the terminal though 🤔

henrikhorluck: Though it is interactive 🤷

FluidSense (approved):
I'm not very well versed in the world of bash scripting, but the rationale looks good to me!

FluidSense:
One note after quickly reading up on some bash lingo: What is the functionality of combining `stderr `and `stdout` to `stdout` (which is what I understood `2>&1` does) before `grep`? Don't you want to exit with an error code if` makemigrations ` fail?

sklirg:
`stderr` and status code (exit code/return code) are different things. You can have `stderr` output even if the process exits with `0` (success), and you can have no `stderr` output if the process exits with non-0.

sklirg:
The reason stderr is sent to stdout is because pipes forward stdout only, and grep works on stdin (which reads the previous stdout), so grep wouldn't have worked on the output into stderr (which I think Python/Django uses for migration output)

FluidSense:
@sklirg I see, that makes sense. Thanks for clarifying for me!

henrikhorluck:
> One note after quickly reading up on some bash lingo: What is the functionality of combining `stderr `and `stdout` to `stdout` (which is what I understood `2>&1` does) before `grep`? Don't you want to exit with an error code if`makemigrations` fail?

Then we would have to set the `pipefail` option https://stackoverflow.com/questions/6871859/piping-command-output-to-tee-but-also-save-exit-code-of-command

PR #2549, opened 2020-10-13
Swagger tag sortation

FluidSense:
## Description of changes

This updates Django Rest Framework and configures OpenAPI-tags to split our Swagger-docs in collapsible sections! 🎉 

## Code Checklist

## Visual changes

<details>
<summary>Before</summary>
  

</details>

<details>
<summary>After</summary>

firefox_qbzQEBZYuJ

</details>

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,41 +1,40 @@
 [[package]]
-category = "main"
-description = "A configurable sidebar-enabled Sphinx theme"
 name = "alabaster"
+version = "0.7.12"
+description = "A configurable sidebar-enabled Sphinx theme"
+category = "main"
 optional = true
 python-versions = "*"
-version = "0.7.12"
 
 [[package]]
-category = "main"
-description = "Low-level AMQP client for Python (fork of amqplib)."
 name = "amqp"
+version = "2.6.1"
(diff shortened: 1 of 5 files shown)


poetry.lock
```diff
@@ -1,41 +1,40 @@
 [[package]]
```
henrikhorluck, line 1: It sadly appears the Poetry-lockfile-format was changed? Could this be related to https://python-poetry.org/blog/announcing-poetry-1-1-0.html? If so we should to the necessary updates to `pyproject.toml`, document the requirement of Poetry `1.1.0`, and probably version poetry-version in our Dockerfiles. We should perhaps also look into wrapping Poetry in our Makefile-setup. Note: much of what I said is beyond the scope of this PR, and should be it's own issue. 
Just check whether `poetry --version` is less than 1.1.0, and we should be good

henrikhorluck (approved):
LGTM, left a comment that should be investigated though

poetry.lock
FluidSense: Yeah, seems I have `1.1.1` locally, and the Docker container had `1.1.2`.

PR #2550, opened 2020-10-14
Sort Swagger tag groups alphabetically instead of folder order

FluidSense:
## Description of changes

Right now, Swagger tag groups are sorted by the name of the folder (inside `apps`) they are found in, which is not visible to the user.

In one way, I kinda like it, as it gives an idea of which folder you can find them in.
Changing the sortation also adds a little pause to the "loading"-screen where the loader freezes as it is finished loading, but the UI is busy sorting it.
However, with no help text which tells you how it's sorted, it looks weird and makes finding things hard. It is also not very necessary for consumers of the API to know the folder name order, if they are not contributing to OW4 as well.

Thus, I propose changing it to alphabetical sort. 

Please feel free to discuss this decision!

**PS**: Swagger UI parameters can be found [here](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/).

### Option 1: Alphabetical
<details>
<summary>Before</summary>
  

</details>

<details>
<summary>After</summary>
  

</details>

### Option 2: Folder name with description
<details>
<summary>Before</summary>
  

</details>

<details>
<summary>After</summary>
  

</details>

### DIFF
--- a/templates/api/swagger-ui.html
+++ b/templates/api/swagger-ui.html
@@ -22,7 +22,8 @@
         requestInterceptor: (request) => {
           request.headers['X-CSRFToken'] = "{{ csrf_token }}"
           return request;
-        }
+        },
+        tagsSorter: 'alpha',
       })
     </script>
   </body>


kharann (approved):
I agree it should being sorted alphabetically. I belive the apps-folder grouping is good enough imo

FluidSense:
@kharann Just a clarification: Endpoints are not sorted by apps-folder, but by their path, e.g. `/api/v1/events/`.
Thus, Committee Application is an example where this can be confusing, as it is located inside the `approval`-app. That is why it was sorted as the first group when using apps-folder name sort which is the current view, since `approval` is the first folder name below `api` where DRF/Swagger is located.

PR #2557, opened 2020-10-18
Update github actions to deploy prod on merge to develop

tokongs:
## Description of changes

It would be nice if we automatically deploy changes to prod when they are merged into develop. 
## Code Checklist

- [ ] I have added tests

## Visual changes

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

### DIFF
--- a/.github/workflows/deploy_prod.yml
+++ b/.github/workflows/deploy_prod.yml
@@ -7,6 +7,17 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          
+      - name: Bump version and push tag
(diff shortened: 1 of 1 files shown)


FluidSense:
@tokongs Do you think we should add a semver-package or something as well, to actually create semver releases and tags? 
Or should we change the release script to just deploy what's on develop?
This change would as of now probably not _actually_ deploy anything, since the deploy scripts looks for new releases.

tokongs:
@FluidSense I think you are right! I will look into how to fix it tomorrow 😎

FluidSense (approved):
Nice! I'm not yet truly a fan of the Angular Commit Message Conventions, but it's better than manual deploys! And it also seems to be configurable if we agree upon another convention

PR #2558, opened 2020-10-18
Update validation method of extra serializer

CarlOfHoly:
Updates validation method of extras to validate single id instead of array of ids


FluidSense (approved):
This reflects the actual workings of the model better!

PR #2559, opened 2020-10-18
Add CodeQL security analysis

henrikhorluck:
## Description of changes

Add CodeQL analysis to CI 

See: https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/about-code-scanning

## Code Checklist

- [ ] I have added tests


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,62 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [develop]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [develop]
+  schedule:
+    - cron: '0 8 * * *'
```
henrikhorluck, line 15:
```suggestion
    - cron: '14 19 * * *'
```

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,62 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [develop]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [develop]
+  schedule:
+    - cron: '14 19 * * *'
```
henrikhorluck, line 15:
```suggestion
```
Remove cron-schedule, since we likely push often enough anyway

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,60 @@
...
+        languages: ${{ matrix.language }}
+        # If you wish to specify custom queries, you can do so here or in a config file.
+        # By default, queries listed here will override any specified in a config file. 
+        # Prefix the list here with "+" to use these queries and those in the config file.
+        # queries: ./path/to/local/query, your-org/your-repo/queries@main
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v1
```
FluidSense, line 46: Do we need to build anything?

henrikhorluck: I don't think so, since we only have JavaScript and Python code

henrikhorluck: I am somewhat unsure about how this interacts with libraries, e.g. Django

FluidSense: I kinda don't think we need to build anything either, since the action itself analyzes the code as far as I understood. It didn't seem to start the entire project and test it out. So perhaps we could make the action faster by removing this step?

henrikhorluck: I removed it now, and we got an error regarding setting up Python dependencies, so I added an export to `Requirements.txt`

tokongs (approved):
LGTM

PR #2560, opened 2020-10-18
OIDC Api management

FluidSense:
## Description of changes

OIDC Clients are the main way we allow users access to data through the API.
Thus, it should be possible to create a management app for OIDC Clients which would allow users to log in and edit configuration for their own clients. 

The OIDC-package that we use (`django-oidc-provider`) sets ID and secret through the admin form. This PR adds these secrets on `Create` and `Update` if the client is `confidential` to allow for a management app to also support creations of new clients.

https://github.com/dotkom/auth-board is meant to be a consumer application for this.

## Code Checklist

- [x] I have added tests
- [x] ~I have provided documentation~ I have provided _some_ documentation. Most of the documentation for this should probably be in the consumer.

## To-Do

- [ ] ~Protect the endpoint with a TokenHasScope (from `oauth2_provider`, oh the irony) to regulate whom can create apps which access sensitive client data.~ I suggest skipping this since very little else is scope protected and the only sensitive information one can access is your own keys.
- [x] Add tests for `get-own`.


FluidSense:
Perhaps we should add scope protection, but as of now, very very few of our apps are using it and our OIDC-provider does not support DRF. I feel that it is thus outside of the scope of this PR

apps/online_oidc_provider/serializers.py
```diff
@@ -31,8 +69,55 @@ class Meta:
...
+        client_secret = ""
+        if validated_data.get("client_type") == "confidential":
+            client_secret = sha224(uuid4().hex.encode()).hexdigest()
+        response_type = validated_data.pop("response_types")
+        data = {
+            **validated_data,
+            "client_id": client_id,
+            "client_secret": client_secret,
+        }
+        print(data)
```
oleast, line 97: Print

oleast (changes requested):
🇰🇿

apps/online_oidc_provider/serializers.py
FluidSense: _damn, you're good_

oleast (approved):

PR #2561, opened 2020-10-19
Allow users to see own permissions

FluidSense:
## Description of changes

If external apps wants to limit access to users without registered permissions, one would now have to do this based on group memberships. This change allows for accessing a users permissions and can thus be used for committee-specific applications without limiting to certain groups, but rather the actual permissions that group has.

**Note:** The endpoint is unable to generate a schema as there is no serializer attached to this view (since permissions are serialized by default). This would perhaps need to manually written as [specified in the docs](https://www.django-rest-framework.org/api-guide/schemas/#map_field). I have added the response type in the text as a temporary fix to this.

### DIFF
--- a/apps/authentication/api/views.py
+++ b/apps/authentication/api/views.py
@@ -1,4 +1,4 @@
-from django.contrib.auth.models import Group
+from django.contrib.auth.models import Group, Permission
 from rest_framework import mixins, status, viewsets
 from rest_framework.decorators import action
 from rest_framework.permissions import AllowAny, IsAuthenticated
@@ -20,6 +20,7 @@ from apps.authentication.serializers import (
     OnlineGroupCreateOrUpdateSerializer,
     OnlineGroupReadOnlySerializer,
     PasswordUpdateSerializer,
+    PermissionReadOnlySerializer,
     PositionCreateAndUpdateSerializer,
     PositionReadOnlySerializer,
     SpecialPositionSerializer,
@@ -57,6 +58,7 @@ class UserViewSet(
         "change_password": PasswordUpdateSerializer,
         "anonymize_user": AnonymizeUserSerializer,
         "dump_data": UserDataSerializer,
+        "user_permissions": PermissionReadOnlySerializer,
     }
 
     @action(detail=True, methods=["put"])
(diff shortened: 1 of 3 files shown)


apps/authentication/api/views.py
```diff
@@ -84,6 +84,28 @@ def dump_data(self, request, pk: int):
         return Response(data=serializer.data, status=status.HTTP_200_OK)
 
 
+class PermissionsViewSet(viewsets.ViewSet):
```
oleast, line 87: Could be very useful for admins to see a users permissions as well.
Could therefore be useful to use a regular modelviewset which takes the user_id as a url param?

FluidSense: Hm, you're right! Perhaps it should rather just be a subpart of `/users/{id}`? As in `/users/{id}/permissions`.

oleast: Since we don't have any support for nesting with this project I think it works best as its own viewset, not under `/users`.  So `user/permissions/{user_id}/` I guess?

My guess is that you could use `user.get_user_permissions()` and turn that into a queryset of Permission somehow?

oleast: > Or perhaps have both, so that you do not need to know your own ID to see your permissions? (although with OIDC, you have your ID from /userinfo)

Any logged in user should be able to get their id, so it would not be much of a problem.
The API-endpoints we have that do this without the user_id are probably ones I created before I knew any better

oleast: Scratch that, I'm mixing things up...
`/users/{user_id}/permissions/` probably makes sense as an `@action` on the user. Some permission class like `IsSelfOrSuperUser`

For listing and editing permissions there should probably be a ModelViewSet.

apps/authentication/api/views.py
```diff
@@ -83,6 +85,44 @@ def dump_data(self, request, pk: int):
...
+        Returns all permissions for a given user.
+        Codename is "action_object", e.g. "delete_profile".
+        Content_type is ID of object type.
+        If you want to group permissions, you can group on object type and map the ID to the object from the codename string.
+        """
+        user = self.get_object()
+        permissions = (
+            Permission.objects.filter(group__user=user)
+            .order_by("content_type")
+            .distinct()
```
oleast, line 100: Won't this just get the permissions assigned to a user through a group?
A user may have permissions assigned to themselves as well?

apps/authentication/api/views.py
```diff
@@ -83,6 +85,44 @@ def dump_data(self, request, pk: int):
...
+        )
+        page = self.paginate_queryset(permissions)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
+        serializer = self.get_serializer(permissions, many=True)
```
oleast, line 106: Could this lead to the response some times including the pagination object (with total count, results array, next page and so on)?
That may trip over a client that always expects the result to be wrapped in that object

FluidSense: That is true! I just figured pagination was a smart thing to keep using when possible since most of the other endpoints does. However, I do agree that you usually want all permissions when you query this endpoint instead of repeated queries until you have reached all pages.

apps/authentication/api/views.py
FluidSense: You're right, trying to find out how I can find all permissions by user while still using the Permission-object.

apps/authentication/api/views.py
oleast: I agree that it could be nice to keep using. How is the `None` page handled for regular ModelViewSets? Is it possible to replicate how they do it, like return an empty paginated response or whatever they do?

apps/authentication/api/views.py
oleast: I think there is a `get_all_permissions`, which also returns strings.

apps/authentication/api/views.py
FluidSense: I'm not sure that ever is a thing to be honest. I've tried with 0 data, but it still becomes a QuerySet with 0 data in the results-section. Just found that error-handling at StackOverflow and thought it might be a thing. But I have never been able to trigger a None-page as of yet.

apps/authentication/api/views.py
FluidSense: Resolved in f4d3847

FluidSense:
I'm not certain how to handle superusers. I originally thought I could just return all permissions, but that doesn't seem to be entirely correct either.
```python
>len(user.get_all_permissions())
516
>len(Permission.objects.all().distinct())
519
```

apps/authentication/api/views.py
FluidSense: Made it a ReadOnlyModelViewSet for now, since editing requires additional logic which is not needed as of now.

tokongs (approved):
LGTM

PR #2564, opened 2020-10-25
Create privacy relation during registration and use privacy settings when creating an Attendee

kharann:
Create a Privacy Object related to user at register rather than when opening profile page.

If you attend an event, and the sent data is undefined, then the photo_consent and allow_visibility_on_attended_event will be set from your Privacy settings


apps/authentication/views.py
```diff
@@ -89,6 +90,9 @@ def register(request):
                 email.primary = True
                 email.save()
 
+                # Set default privacy settings
+                Privacy.objects.create(user=user)
+
```
oleast, line 95: Users can also sign up via the api (POST to `/api/v1/users/`), which this would not cover.

What about doing this as a `post_save` signal on OnlineUser when the `created` flag is `True`? That would cover both cases 😄

apps/authentication/appconfig.py
```diff
@@ -18,3 +24,4 @@ def ready(self):
         watson.register(
             OnlineUser, fields=("first_name", "last_name", "ntnu_username", "nickname")
         )
+        post_save.connect(set_default_privacy, sender=User)
```
oleast, line 27: I'm a bit unsure on how signals work, but maybe it should be `OnlineUser`, there is something about them not reacting to subclasses

apps/authentication/signals.py
```diff
@@ -130,6 +137,12 @@ def set_staff_status(sender, instance: GroupMember, created=False, **kwargs):
     user.save()
 
 
+@receiver(post_save, sender=User)
+def set_default_privacy(sender, instance: User, created=False, **kwargs):
```
oleast, line 141: I'm a bit unsure on how signals work, but maybe it should be `OnlineUser`, there is something about them not reacting to subclasses

s0lvang:
As we have had bugs related to this i would consider writing a few tests :) 

image

apps/authentication/views.py
sklirg: Like this? 😏 https://github.com/dotkom/onlineweb4/blob/develop/apps/profiles/signals.py#L10

kharann: I looked a little and it and I think that signal might be a bit buggy because it doesn't work for new users until they click My Page. Might be that the Signal had to be connected to OnlineUser

sklirg: "min side" view has a hack which does this, so that might be the case. I'm positive it _has_ worked before though, as we had problems with this before.

apps/profiles/models.py
```diff
@@ -22,9 +22,7 @@ class Privacy(models.Model):
         _("greit å ta bilder på arrangement"), default=None, null=True
     )
 
-    user = models.OneToOneField(
-        User, related_name="privacy", on_delete=models.CASCADE, primary_key=True
-    )
+    user = models.OneToOneField(User, related_name="privacy", on_delete=models.CASCADE)
```
oleast: Why remove the primary key here?

kharann: I'm just trying out a few different things. Privacy in a signal actually screws up a bunch of things (some transaction and atomic blocs). Do basically just a WIP until I figure out an alternative to postsave

apps/authentication/models.py
```diff
@@ -313,7 +313,10 @@ def image(self):
 
     def get_visible_as_attending_events(self):
         """ Returns the default value of visible_as_attending_events set in privacy/personvern """
-        if hasattr(self, "privacy"):
```
henrikhorluck, line 316: It is okay to remove this `hasattr`-check, since we create privacy for everyone with the migration

apps/profiles/migrations/0009_auto_20201106_1816.py
```diff
@@ -0,0 +1,38 @@
...
+import django.db.models.deletion
+
+
+def create_privacy_for_all(apps, schema_editor):
+    # we used to have that an user could not have a privacy,
+    # this creates it for everyone
+    User = apps.get_model("authentication", "OnlineUser")
+    Privacy = apps.get_model("profiles", "Privacy")
+    for user in User.objects.filter(privacy=None):
+        Privacy.objects.create(user=user)
```
henrikhorluck, line 14: This should work @kharann ?

kharann: Yup!

henrikhorluck (approved):
LGTM :shipit:

henrikhorluck:
This is clear to merge @kharann ?

PR #2567, opened 2020-11-13
changed from group.name to group.long_name in serializer

s0lvang:
Fixing dump data continued

### DIFF
--- a/apps/authentication/api/serializers/user_data.py
+++ b/apps/authentication/api/serializers/user_data.py
@@ -140,7 +140,7 @@ class WikiAttachmentRevisionSerializer(serializers.ModelSerializer):
 
 class WikiArticleSerializer(serializers.ModelSerializer):
     current_revision = WikiArticleRevisionSerializer()
-    group = serializers.CharField(source="group.name")
+    group = serializers.CharField(source="group.name_long")
 
     class Meta:
         model = WikiArticle


FluidSense (approved):
Just try 👌

apps/authentication/api/serializers/user_data.py
```diff
@@ -140,7 +140,7 @@ class Meta:
 
 class WikiArticleSerializer(serializers.ModelSerializer):
     current_revision = WikiArticleRevisionSerializer()
-    group = serializers.CharField(source="group.name")
+    group = serializers.CharField(source="group.name_long")
```
oleast, line 143: I think this group is a Django group, which doesn't have a long name?

FluidSense: The error says that it doesn't have a name either, which is why my comment is "Just try" 😅 😂 Neither is correct, any idea of what it should be @oleast ?

oleast: It's supposed to be group :/

https://github.com/django-wiki/django-wiki/blob/master/src/wiki/models/article.py#L68

FluidSense: Hm, seems like the actual issue is that it is a nullable field, which then crashes on "None.name" when there is no group.
https://www.django-rest-framework.org/api-guide/fields/
Perhaps one could pass the parameter "default" with argument "Ingen gruppe", to handle such cases.

PR #2574, opened 2021-02-07
Redirect Dataporten study confirmation to OWF if request started from OWF

FluidSense:
## Description of changes

To allow for calls from OWF to add membership through dataporten, we can just see if the original request comes from OWF, then redirect to OWF again on finished Authorization Flow.

## Code Checklist

- [ ] I have added tests


CarlOfHoly (approved):
LGTM!

PR #2581, opened 2021-03-16
Add option to filter OnlineGroups based on a given user id

kharann:
This Pull request add the possibility to filter OnlineGroups (api/v1/online_group) based on an user_id given in the url

### DIFF
--- a/apps/authentication/api/filters.py
+++ b/apps/authentication/api/filters.py
@@ -1,7 +1,7 @@
 import django_filters
 from watson import search as watson_search
 
-from ..models import OnlineUser
+from ..models import OnlineUser, OnlineGroup, GroupMember
 
 
 class UserFilter(django_filters.FilterSet):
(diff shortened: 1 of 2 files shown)


apps/authentication/api/filters.py
```diff
@@ -15,3 +15,15 @@ def watson_filter(self, queryset, name, value):
     class Meta:
         model = OnlineUser
         fields = ("first_name", "last_name", "rfid", "query")
+
+
+# This is a filter for finding which groups an user is a member of
+def filter_member_user(_, __, value):
+    members = GroupMember.objects.filter(user=value)
+    return OnlineGroup.objects.filter(members__in=members)
```
oleast, line 23: This works, but I'm afraid we're double (or maybe triple) tapping the database with this method?

I think we might be able do do this directly with a simple query like this:
``` python
class OnlineGroupFilter(django_filters.FilterSet):
    members__user = django_filters.NumberFilter(field_name="members", lookup_expr="user_id")

    class Meta:
        model = OnlineGroup
        fields = ("name_short", "name_long", "parent_group", "group_type")
```

Issue #2576, opened 2021-02-22
Migrate from unmaintained OIDC library (django-oidc-provider)


henrikhorluck:
As per https://github.com/juanifioren/django-oidc-provider/issues/377, `django-oidc-provider` is unmaintained and we should look into finding a replacement asap

FluidSense:
Just waiting for this one since we already depend on it for oauth2 https://github.com/jazzband/django-oauth-toolkit/pull/915

FluidSense:
_In progress_

### Critical tasks:
_These are the tasks that needs to be done before PR can be created_

- [ ] Allow `openid`-scope
- [ ] Get `id_token` as JWT  
- [ ] Check that existing scopes do not break
- [ ] Simplify existing scope names
- [ ] Add new scopes
- [ ] Readd extended Application Model 
- [ ] Add connected apps in `dump_data`
- [ ] Add connected apps for OWF-"Personvern"-view.
- [ ] Style authorization-page
- [ ] Replace DRF-endpoints for [Authentication Dashboard](https://github.com/dotkom/auth-board)
- [ ] Ensure that users can see currently active `refresh token`s and revoke them in case they are stolen.

### Descriptive tasks
_Various fields could be implemented for us admins to easier see what a client is used for_

- [ ] Add description as nullable field on Application model
- [ ] Add hosted website url


### Replacement tasks
_I suggest we give a week or two of migration period, where both `django-oidc-provider` and `django-oauth-toolkit` exists alongside each other so that people can update their applications_

- [ ] Remove endpoints for `django-oidc-provider`
- [ ] Remove `django-oidc-provider`-fields from `admin`
- [ ] Remove connected `django-oidc-provider`-apps from `dump-data`
- [ ] Remove connected `django-oidc-provider`-apps from OWF-"Personvern"-view.

FluidSense:
## Changelog 🆕 

To easier write the e-mail to client owners, I will write down breaking changes here as I work.

### Scopes 🔭 
 - The scope `onlineweb4` is renamed. This scope used to add claims to /userinfo, with different information regarding your user, such as address, username, etc. This is renamed to `profile`. Please change to using `profile`.  
 
### URLs 💻 
 - Previous urls were prefixed with `openid/`. The new prefix is `sso/`.

FluidSense:
#  Add connected apps for OWF-"Personvern"-view.

[...]

PR #2620, opened 2021-04-18
Enable OIDC-use of django-oauth-toolkit

FluidSense:
## Description of changes

This is the first step of refactoring to 1 unified oauth2/oidc-provider instead of 2. #2576 🎉 
`django-oauth-toolkit` is more maintained and has better support for django rest framework (e.g. scope endpoint protection).

This also changes the old "External app"-page to a new one, which is intended to have less information to ensure that the user focus on the actual content, and hopefully reads what they give away of data. 🤓 🎨 

With this, we want users to move their clients over from the previous system to the new one. 
The old ones will no longer be accessible in authentication dashboard when the related PR is merged.

NB!:
Certain apps using our oauth2-clients _might_ be impacted by any change here, but it only extends, not modifies, existing behaviour, so fingers crossed 🤞 

## Code Checklist
- [ ] I have added tests
  - Some documentation is in the description of the endpoints, while more documentation is intended to be part of the authentication dashboard app.

## Deploy Checklist
 - [ ] Add an RSA-key on Nansen
 
 Set RSA-key through:
 
 ```sh
 openssl genrsa -out oidc.key 4096
 export OW4_OIDC_RSA_PRIVATE_KEY=`cat ./oidc.key`
 ```

## Visual changes

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>
  

image

</details>


FluidSense:
Some `less`-linting fails as my Prettier-config disagree. Have to disable auto-format to fix it.
A bit confused by `Expected double colon pseudo-element notation          /    selector-pseudo-element-colon-notation` rule, as I can't get the less-rule to work as intended if I change to double-colon

FluidSense:
A lot of the tests fail on `401 != 403`, not sure if the expected HTTP response was changed in the new `django-oauth-toolkit`-version or if the authentication uses the wrong protocol. Gotta investigate 🕵️

apps/sso/permissions.py
```diff
@@ -0,0 +1,26 @@
...
+class IsOwnerOrSuperUser(permissions.BasePermission):
+    def has_object_permission(self, request, view, obj):
+        return request.user.is_authenticated and (
+            request.user.is_superuser or request.user == obj.user
+        )
+
+
+class TokenHasScopeOrSuperUser(permissions.BasePermission):
+    def has_object_permission(self, request, view, obj):
+        print("user: ", request.user)
```
oleast, line 13: ➖

apps/sso/serializers.py
```diff
@@ -0,0 +1,120 @@
...
+    get_access_token_model,
+    get_application_model,
+    get_grant_model,
+    get_refresh_token_model,
+)
+from rest_framework import serializers
+
+from apps.sso.models import ApplicationConsent
+
+# TODO:
```
aslakhol, line 12: I think you should sort this out before merging.

aslakhol (approved):
This is large and virtually impossible to do real validation through a PR. That being said, I spent some time walking through the code and I can't see anything I strongly dislike or that seems wrong. 

LGTM

apps/sso/serializers.py
```diff
@@ -0,0 +1,120 @@
...
+# Remove oauth2 from names, find a better naming
+
+
+class Oauth2ClientNonSensitiveSerializer(serializers.ModelSerializer):
+    client_type = serializers.ChoiceField(
+        choices=["public", "confidential"], allow_blank=True, default="public"
+    )
+    algorithm = serializers.HiddenField(default="")
```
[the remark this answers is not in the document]
FluidSense: Empty string is a valid default value actually! Empty string = no algorithm = no OIDC support.

apps/sso/serializers.py
```diff
@@ -0,0 +1,120 @@
...
+        instance.save()
+        return instance
+
+
+class Oauth2ClientSerializer(serializers.ModelSerializer):
+    redirect_uris = serializers.SerializerMethodField()
+
+    class Meta:
+        model = get_application_model()  # Allows for swappable application model
+        fields = "__all__"
```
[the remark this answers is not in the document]
FluidSense: Hm, you're right! This serializer is only meant for endpoints that are capable of handling sensitive information, and at the moment the endpoint both requires a certain scope and that a user can only see their own clients' information. However, we might still want to have the full list instead so that it's obvious what kind of information we give out. And perhaps a name that indicates how "dangerous" it might be

apps/sso/serializers.py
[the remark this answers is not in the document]
oleast: Ah, thanks for the explanation!

apps/sso/serializers.py
FluidSense: You're right. Any ideas for name?

FluidSense:
> This pull request introduces 1 alert when merging 1ddbfd4 into 1dd5ebc - view on LGTM.com
>
> new alerts:
>
>   1 for Call to eval-like DOM function

Is this alert coming from 1dd5ebc? I cannot see how 1ddbfd4 touches anything related to the warning

templates/sso/sso_base.html
```diff
@@ -0,0 +1,68 @@
...
+            crossorigin="anonymous">
+        </script>
+        <script>Sentry.init({ dsn: '{{ sentry_dsn }}' })</script>
+        <script
+            src="https://code.jquery.com/jquery-2.2.4.min.js"
+            integrity="sha256-BbhdlvQf/xTY9gja0Dq3HiwQF8LaCRTXxZKRutelT44="
+            crossorigin="anonymous"
+        ></script>
+        <script>
+            window.jQuery || document.write('<script src="{{ STATIC_URL }}js/libs/jquery-2.2.4.min.js"><\/script>');
```
henrikhorluck, line 47: @FluidSense this is the line triggering the security-warning from LGTM, the commit it is referring to is just the latest commit in `main`

FluidSense: I see, thanks for clarifying! That is just the regular base-html stripped of the nav-bar and footer, I believe I should be able to remove jquery as I am not aware of anything using it on the authorization-page!

apps/sso/serializers.py
FluidSense: Perhaps `ExternalAuthentication` as prefix?

apps/sso/serializers.py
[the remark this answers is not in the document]
FluidSense: Perhaps I should add a comment on it for the next person looking at it!

apps/sso/serializers.py
FluidSense: Went with `SSO`!

FluidSense:
> A lot of the tests fail on 401 != 403, not sure if the expected HTTP response was changed in the new django-oauth-toolkit-version or if the authentication uses the wrong protocol. Gotta investigate 🕵️

I _believe_ it was due to the update instead of change of package, but if it was the latter, it will be OK once everything is moved to `django-oauth-toolkit` anyway

PR #2631, opened 2021-04-26
Don't create consent if no user in authorization

FluidSense:
If authorization happens through a method which is not connected to a user, do not try to create a consent because none were given. 
E.g. when Nibble logs in a user, it is all through `client_credentials` with no user-approval.


kharann (approved):
Lgtm :)))

PR #2632, opened 2021-04-26
Load RSA-key from file instead of env

FluidSense:
OW4-deployment does not execute the `.env`, and as the file is large, it would clutter the file to have the entire content in env. Therefore, we load from file instead.

### DIFF
--- a/onlineweb4/settings/base.py
+++ b/onlineweb4/settings/base.py
@@ -64,12 +64,22 @@ SLACK_INVITER = {
 }
 
 # SSO / OAuth2 settings
+
+OIDC_RSA_PRIVATE_KEY = "" # Default case
+if not os.path.isfile("oidc.key"):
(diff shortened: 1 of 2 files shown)


onlineweb4/settings/base.py
```diff
@@ -64,12 +64,22 @@
...
+
+OIDC_RSA_PRIVATE_KEY = ""  # Default case
+if not os.path.isfile("oidc.key"):
+    print("Failed to locate OIDC RSA Private key file during setup.")
+else:
+    with open("oidc.key", "r") as f:
+        OIDC_RSA_PRIVATE_KEY = f.read()
+
+print("KEY: ", OIDC_RSA_PRIVATE_KEY)
+
```
henrikhorluck, line 76:
```suggestion
```

henrikhorluck (approved):
LGTM

PR #2744, opened 2021-09-19
Add a nested route on OnlineGroup to retrieve user info from the members

kharann:
## Description of changes

Added a route on `online-groups/{group}` where you can get the user information on the members


apps/authentication/api/views.py
```diff
@@ -185,8 +185,16 @@ class OnlineGroupViewSet(MultiSerializerMixin, viewsets.ModelViewSet):
     serializer_classes = {
         "write": OnlineGroupCreateOrUpdateSerializer,
         "read": OnlineGroupReadOnlySerializer,
+        "user_members": UserReadOnlySerializer,
     }
 
+    @action(detail=True, methods=["get"], url_path="user_members")
+    def user_members(self, request, pk: int = None):
```
jfremstad, line 192: Shouldn't this be named group_members?

PR #2759, opened 2021-10-31
Simplify Django Wiki-setup

henrikhorluck:
- Removes customization
- Remove tree-view which currently only gives timeout
- Mostly a POC

## Code Checklist

- [ ] I have added tests

## Visual changes

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>


henrikhorluck:
The primary issue with upgrading and keeping our current HTML-setup is that we would then have both Bootstrap 3 and 4 active on the same page, which is just flat out messy

PR #2859, opened 2022-03-29
Update mark duration from 30 to 20 days

Thomhas:
## Description of changes

Online wishes to change the duration of a mark from 30 days to 20 days. 
## Code Checklist

- [ ] I have added tests

## Visual changes

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>


henrikhorluck:
It might be preferable to make the duration configurable instead, and make a db-migration so that it defaults to 30 for existing marks, so that this does not have a retroactive side-effect

Thomhas:
The change on the mark duration should affect all marks(not only the new ones). With that in mind is it still preferable to attach the duration to the MarkRuleSet?

anderssonw:
status? @Thomhas 😳😳😳😳😳

Thomhas:
> status? @Thomhas 😳😳😳😳😳

Soon!

kharann (approved):
hacky but good as we change mark rules once every 5 years

PR #2929, opened 2022-08-19
Set title max length to 200 on notificaitons

kharann:


henrikhorluck:
This needs a db migration. Will close it for now, should be super easy to recreate

PR #2953, opened 2022-10-26
Add is_active field to username serializer

Terbau:
## Description of changes
The change is needed to be able to see if a user is active or not whenever you request group members.


Terbau:
Nevermind, the field wasn't needed afterall.

Issue #2887, opened 2022-05-09
Billed for wrong items in webshop


henrikskog:
**Describe the bug**
People are being billed for items not chosen when buying items. One of ways in which the bug can occur:  selecting an item, then unselecting it. If you then go to buy something else, you will also be billed for the item you selected then unselected. 

There has been multiple bug reports for this. This is the latest one:

“Heisann! Det kan virke som problemet med overbetaling om man fjerner en artikkel fra webshop-handlekurven fortsatt ikke er borte :sweat_smile: Er en som fjernet en sløyfe når vedkommende skulle kjøpe ølglass og ble alikevell fakturert for sløyfen i stripe, men registrert for kun ølglassene :see_no_evil:”

**To Reproduce**

**Expected behavior**
- Users should only be billed for the items that are in their cart when pressing buy!!

**Screenshots**

marcelwk:
#### To reproduce
- Go to webshop and add two items into cart
- Remove one of them
- Check the active orderline (https://old.online.ntnu.no/api/v1/webshop/orderlines/)

Notice how `Orderline.subtotal` (which is shown in the UI) shows the price of the remaining item, while `Orderline.payment.payment_prices.price` (which is sent to Stripe) shows the price of both items.

Also, if you add a third item after the removal, both values become correct.

Seems like the issue is that the payment object is only updated when a new item is added, and not on removal.

PR #2978, opened 2023-01-22
make payment price sync on order delete DOT-263

marcelwk:
## Description of changes
Fix of #2887 

Added decorator to make price sync function also when order is deleted, and not only when new order is created.
Otherwise the price shown in UI and price sent to Stripe diverge after order is deleted

## Code Checklist

- [ ] I have added tests

### DIFF
--- a/apps/webshop/signals.py
+++ b/apps/webshop/signals.py
@@ -1,6 +1,6 @@
 import logging
 
-from django.db.models.signals import post_save
+from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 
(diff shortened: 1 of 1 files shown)


marcelwk:
Have not been able to set up the dev-env to test the fix. I would be grateful if someone could test it, and optionally create a test for the order-removal case.

marcelwk:
@henrikskog, ping because you're the issue author

henrikskog:
Sorry for reviewing so late, but thank you so much for the PR! I ran it locally and as far as I can tell it now syncs correctly.

henrikskog (approved):

PR #2989, opened 2023-02-06
Fix dotkom possible duplicate address in error mail

Terbau:
## Description of changes
- Fixed a possible issue where dotkom's email address could potentially be included twice when sending an error email.


kharann (approved):
GGGGG

PR #2996, opened 2023-02-09
Create boto3 client from new session

Terbau:
## Description of changes
- Attempts to fix the error `KeyError: 'endpoint_resolver'` by creating a boto3 client from a session each time rather than sharing a session between threads.

## Code Checklist

- [ ] I have added tests


henrikhorluck:
Is this still relevant?

PR #3005, opened 2023-03-18
Change default of feedback gives mark to False 

Terbau:
Temporarily change the default value of gives_mark to False. This is because there is an issue with mails not always being delivered which results in unwanted marks (prikker) being given out.


henrikhorluck:
Rather fix the root issue

PR #3056, opened 2023-11-04
feat: add `PaymentTypes` enum

brage-andreas:
* Added check to account for `payment.deadline`

This PR fixes an issue where a payment with type delay will ignore a payment's deadline, and potentially set a deadline earlier than a user might expect.


brage-andreas:
I figured I probably need to alter lines: [view_tests.py#L203-L253](https://github.com/dotkom/onlineweb4/blob/24a4426f6020248b26f28d9e43ae1abb33ef1a4f/apps/payment/tests/view_tests.py#L203-L253), but I am unsure if this is needed nor am I sure exactly how.

apps/payment/models.py
```diff
@@ -48,7 +52,7 @@ class Payment(models.Model):
     )
     """Which Stripe key to use for payments"""
 
-    payment_type = models.SmallIntegerField(_("type"), choices=TYPE_CHOICES)
+    payment_type = models.SmallIntegerField(_("type"), choices=PaymentTypes.choices)
```
henrikhorluck, line 55: This requires a db migration (that does nothing). Run `./manage.py makemigrations` (needs to be formatted, but should be handled by the pre-commit hook if you have it set up)

apps/events/utils.py
```diff
@@ -71,8 +71,12 @@ def _handle_waitlist_bump_payment(payment, attendees):
                 "Dette arrangementet krever betaling og du har 48 timer på å betale"
             )
 
-    elif payment.payment_type == 3:  # Delay
+    elif payment.payment_type == PaymentTypes.DELAY:  # Delay
         deadline = timezone.now() + payment.delay
+
+        if payment.deadline:
+            deadline = max(deadline, payment.deadline)
+
```
henrikhorluck, line 79: This needs to be handled more places than just the waitlist bumping. We should probably change `create_payment_delay` directly, but ensure that it still behaves the same for `PaymentTypes` that are not `utsettelse`

I think the calculated deadline is used later on so the method should probably return the `PaymentDelay` it creates

apps/events/utils.py
```diff
@@ -71,8 +71,12 @@ def _handle_waitlist_bump_payment(payment, attendees):
                 "Dette arrangementet krever betaling og du har 48 timer på å betale"
             )
 
-    elif payment.payment_type == 3:  # Delay
+    elif payment.payment_type == PaymentTypes.DELAY:  # Delay
```
henrikhorluck, line 74: You can remove the comments, since the code is now self-documenting

henrikhorluck (changes requested):
> I figured I probably need to alter lines: [view_tests.py#L203-L253](https://github.com/dotkom/onlineweb4/blob/24a4426f6020248b26f28d9e43ae1abb33ef1a4f/apps/payment/tests/view_tests.py#L203-L253), but I am unsure if this is needed nor am I sure exactly how.

You should not need to change the mommy tests—we are not changing how mommy works, just setting a different deadline in one specific case. 

We might want to test that `create_payment_delay` (if that is where we change the logic) gives the "right" deadline since we are now adding logic to it.

apps/events/utils.py
henrikhorluck: Okay I stand somewhat corrected. If the payment type is `DEADLINE`, and the person is bumped from the waitlist with less than 2 days to the deadline (or when the deadline has passed), they already get at least a two day `payment_delay`. I was thinking about the case where you register _directly_, without getting bumping, which should behave the same (if the payment delay has passed), but you currently just get an expired deadline, and get kicked out once mommy runs.

In other words: we don't need to change anything, just use `DEADLINE`. 

Signing up directly when the payment deadline has passed (`handle_attend_event_payment`), and the payment type is `DEADLINE`, it should probably behave like getting bumped

henrikhorluck (approved):
LGTM :shipit:

PR #3094, opened 2024-01-03
Fix sso consents integrity error

Terbau:
Whenever someone attempted to use the sso authorization endpoint with an authorization code, it would return a 500 error after the first time since it couldn't update the existing sso consents for that client and user.

### DIFF
--- a/apps/sso/signals.py
+++ b/apps/sso/signals.py
@@ -10,9 +10,16 @@ def handle_app_authorized(sender, request, token, **kwargs):
         return
     if not token.user:
         return
-    ApplicationConsent.objects.create(
-        user=token.user, client=token.application, approved_scopes=token.scopes
+
+    consent, created = ApplicationConsent.objects.get_or_create(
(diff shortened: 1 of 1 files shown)


henrikhorluck (approved):
I would highly recommend to always write a failing test case before you fix a bug. 

This appears to have been "fixed" earlier in https://github.com/dotkom/onlineweb4/commit/a39e24c9f8e969d55413feccdc6f585f83c2db87.
It is not clear to me why the `token.application.pk` should be the `pk` of the `ApplicationConsent` and not the pk of the client? Probably a bug, since the integrity check is `user + client`. Then the old fix should've been:

```diff
if ApplicationConsent.objects.filter(
-        pk=token.application.pk, user=token.user
+        client=token.application, user=token.user
    ).exists():
```

And then this PR adds support for updating the token scopes of an existing application. I think you can probably remove the first `if`, as it is now fine if it exists.

I find it strange that this has not been found earlier.

PR #3118, opened 2024-01-25
Add Cognito export and just-in-time pwd-setting

henrikhorluck:
Co-Authored-By: Henrik Skog <henrikskog01@gmail.com>
Co-Authored-By: Mats Jun <mats@jun.codes>

Needs config in vault before deploy to prod, which means we need to set `OW4_COGNITO_USER_POOL_ID` and give more permissions in `zappa_settings.yml` (under `extra_permissions`)? IIRC that was initially set up in the terraform monorepo, [here](https://github.com/dotkom/terraform-monorepo/blob/8c6c200689e0106155dcd3fe2f259b96781679c2/modules/zappa-deploy-actions/main.tf#L15), I would guess we need to update both places?


junlarsen (approved):
Yep we're going to need toadd cognito actions to iam. I think the scopes we need are:

- `cognito-idp:AdminSetUserPassword`
- `cognito-idp:AdminCreateUser`
- and perhaps `cognito-idp:AdminGetUser`

PR #3121, opened 2024-01-26
Support Cognito JWT Auth in backend, remove django-oauth-toolkit

henrikhorluck:
- Remove weird migration dependency
- Remove shop and inventory aka nibble
- Remove django-oauth-toolkit and the "new" OIDC-idp
- Support auth to backend with Cognito JWT

I have verified locally that this does work with vengeful vineyard (with a patch to use Cognito as Identity provider, thus supplying a Cognito-signed JWT), and OWF using our django-oidc-provider IDP in OW4

# What this breaks

It breaks all existing uses of the `/sso`-based OIDC-clients authenticating towards OW4, which should be essentially no-one, since config for that was completely broken for the last couple years, anyways.

This also breaks OAuth2-only uses of the same API, which I think was only used for nibble? Which is no longer in use.


henrikskog (approved):
Awesome, good job! 🤩 🤩 I trust you in determining when it is ready.

PR #3136, opened 2024-02-20
Store event user actions to gain insights form it in the future

henrikskog:


henrikhorluck (commented):
A probably easier way to implement this (given ow4 has so many ways to attend/unattend) is to use https://docs.djangoproject.com/en/5.0/topics/signals/, and listen to create/delete on Attendee

henrikskog:
Don't wanna learn about that Henrik :( I'm 99% sure I covered everything here so should be gucci

junlarsen:
postgres triggers? :D

henrikskog:
Okay you were right @henrikhorluck, worked great using signals. I have ran this locally and it is working perfectly for attendance through old and the dashboard. I have not been able to get the frontend up and running against the backend, so I have not tested that but it should work. This is a low risk PR seeing that it just saves stuff to a new table, so I would like it if we could just merge to test in dev

apps/events/signals.py
```diff
@@ -0,0 +1,18 @@
...
+from django.dispatch import receiver
+
+from .models import Attendee, EventUserAction
+
+
+@receiver(signal=pre_save, sender=Attendee)
+def handle_payment_relation_status_change(sender, instance: Attendee, **kwargs):
+    EventUserAction.objects.create(
+        user=instance.user, event=instance.event.event, type="register"
+    )
```
henrikhorluck, line 11: This should rather use post_save and check that `created` is true https://docs.djangoproject.com/en/5.0/ref/signals/#post-save

henrikskog: True that would have been better

PR #3138, opened 2024-02-21
show allergies on users at waitlist in dashboard events

madsab:
## Description of changes
- Add "allergies" section to users on waitlist in dashboard events
   - **REASON:** Bedkom wants to know allergies of users on waitlist so that they are prepared if the user becomes an attendee

## Code Checklist
- [x] Add allergies to users on waitlist in dashboard events


henrikhorluck (approved):
LGTM

PR #3167, opened 2024-04-17
Add wine penalty manager role

Terbau:
This is needed in order for vengeful vineyard to auto sync the role to the user.


brage-andreas (approved):
Looks good!

PR #3169, opened 2024-04-21
fix email sending bug

henrikskog:
Mail sending in the current version of django has a bug with some versions of python. 3.11.8 should be compatible.

See https://github.com/django/django/pull/17979#issuecomment-2063271301

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ COPY --from=js-static $APP_DIR/bundles ./bundles
 
 RUN ./manage.py collectstatic
 
-FROM amazon/aws-lambda-python:3.11.8
+FROM amazon/aws-lambda-python:3.11.2024.03.28.16
 
 LABEL maintainer="Dotkom <dotkom@online.ntnu.no>"
 ENV POETRY_VIRTUALENVS_CREATE=false \
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ COPY --from=js-static $APP_DIR/bundles ./bundles
 
 RUN ./manage.py collectstatic
 
-FROM amazon/aws-lambda-python:3.8
+FROM amazon/aws-lambda-python:3.11.8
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -41,7 +41,7 @@ COPY --from=js-static $APP_DIR/bundles ./bundles
 
 RUN ./manage.py collectstatic
 
-FROM amazon/aws-lambda-python:3.11
+FROM amazon/aws-lambda-python:3.8
```
henrikhorluck: 3.11.8, not 3.8

henrikhorluck (changes requested):

Dockerfile
henrikskog: yep fixed

henrikhorluck (approved):

Dockerfile
```diff
@@ -41,7 +41,7 @@ COPY --from=js-static $APP_DIR/bundles ./bundles
 
 RUN ./manage.py collectstatic
 
-FROM amazon/aws-lambda-python:3.8
+FROM amazon/aws-lambda-python:3.11.8
```
henrikhorluck, line 44: 3.11.8 is not a tag AWS publishes https://hub.docker.com/r/amazon/aws-lambda-python/tags?page=&page_size=&name=3.11&ordering=last_updated. Python [3.11.9](https://pythoninsider.blogspot.com/2024/04/python-3119-is-now-available.html) (which caused this bug) was released april 8th

[3.11.2024.03.28.16](https://hub.docker.com/layers/amazon/aws-lambda-python/3.11.2024.03.28.16/images/sha256-0e24f7bb734e2d668406e08be78b38bb3c08522cd94c242e8275f4a8533ec410?context=explore) is the tag from before 3.11.9 launch
```suggestion
FROM amazon/aws-lambda-python:3.11.2024.03.28.16
```

henrikhorluck (changes requested):

PR #3190, opened 2024-06-06
Support new mark system

henrikhorluck:
Marihøner har prikker. 🐞 

Rules implemented: https://old.online.ntnu.no/wiki/komiteer/nyeprikkregler/

Tests should be re-calculated to verify that the new values are correct. This also needs more tests covering the example cases from the rules


henrikskog (approved):
Great job at implementing this! Let's get it into the staging env and do some testing :)

PR #3207, opened 2024-07-27
Set name in auth0 as well

henrikskog:


henrikhorluck:
We already override `save` in `OnlineUser` — this seems unnecessary

PR #3210, opened 2024-07-28
Add database cache and cache attendees

henrikskog:


apps/events/api/views.py
```diff
@@ -155,10 +156,18 @@ def unregister(self, request, pk=None):
         url_path="public-attendees",
     )
     def public_attendees(self, request, pk=None):
+        CACHE_TIME = 10
+        sentinel = object()
+        cached = cache.get("public-attendees" + str(pk), sentinel)
+        if cached is not sentinel:
+            return cached
+
```
henrikhorluck, line 164: Much less code if you use `@method_decorator(cache_page(10))` https://www.django-rest-framework.org/api-guide/caching/#using-cache-with-apiview-and-viewsets

onlineweb4/settings/django.py
```diff
@@ -23,6 +23,13 @@
 # might be 100% safe to remove, but keep it safe
 ATMOIC_REQUESTS = True
 
+CACHES = {
+    "default": {
+        "BACKEND": "django.core.cache.backends.db.DatabaseCache",
```
henrikhorluck, line 28: If you use the DatabaseCache I don't think this actually speeds anything up?

henrikhorluck (changes requested):

onlineweb4/settings/django.py
henrikskog: Some db queries may be extra slow so it's worth caching, no? 👍

apps/events/api/views.py
henrikskog: Nice!

onlineweb4/settings/django.py
henrikskog: I see what you mean though. But this requires no further infra, easy to implement and it may have some gains at least

henrikhorluck: Yes, but this is caching it _in the database_ making it probably almost as slow as running the query itself, you can make a performance-test with locust. I have [scaffolding](https://github.com/dotkom/onlineweb4/commit/fa06eb27577eeb7feb6c64d6870b5abb93bee691) from a previous test

henrikskog: Hmm so you're saying that there are no queries so slow that they're worth caching in the same db. You may be right, I haven't done extensive testing. I thought it would be worth it for queries with a lot of joins but the gains may be negligible compared to the network delay. 

This is only a draft btw

henrikhorluck: There are definitively queries worth caching / not performing, and we _should_ have caching like this in most places.

Thanks to sentry we actually have quite a bit of insight into this, so we do actually know what causes the most issues: https://dotkom.sentry.io/profiling/?project=204971&statsPeriod=90d

Essentially it's mostly the event-api

PR #3216, opened 2024-08-02
Remove ntnu username requirement on membership approval

henrikskog:


apps/approval/dashboard/views.py
```diff
@@ -104,7 +93,7 @@ def approve_application(request):
                 membership.save()
             else:
                 membership = Membership()
-                membership.username = user.ntnu_username
+                membership.username = user.email
```
henrikhorluck, line 96: This should be user-id if it is something that isn't the email, but this is actually a bit of a assumption-error / issue that we might have been giving out memberships to new master-students without NTNU-accounts 🤔

henrikhorluck (changes requested):

henrikhorluck:
Superceded by #3218

PR #3218, opened 2024-08-02
Revert form asking for userinfo and set name from FEIDE

henrikhorluck:
Review commit by commit

This fixes the issue of being unable to apply for master if you previously have a bacheor-membership, which was a bug introduced in #3205 

End resultat: user needs to authenticate against FEIDE, before a manual approval, and we then get the information we need.

Not supported: new master members (from non-NTNU) signing up for ITEX before they get a NTNU-account. I don't think we should support that.

### DIFF
--- a/apps/approval/urls.py
+++ b/apps/approval/urls.py
@@ -14,11 +14,6 @@ urlpatterns = [
         name="approval_send_membership_application",
     ),
     re_path(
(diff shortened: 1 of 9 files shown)


jotjern (approved):
LGTM, think we need to fix this a different way.

PR #3228, opened 2024-08-15
Add Online identifier to calendar event titles

jorgengaldal:
## Description of changes

Add identifier to calendar event titles to make it easier to differentiate Online's events from other events in one's calendar.

## Code Checklist

- [ ] I have added tests

## Visual changes

<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

### DIFF
--- a/apps/events/utils.py
+++ b/apps/events/utils.py
@@ -161,7 +161,7 @@ class EventCalendar(Calendar):
         cal_event.add("dtstart", event.event_start)
         cal_event.add("dtend", event.event_end)
         cal_event.add("location", event.location)
-        cal_event.add("summary", event.title)
+        cal_event.add("summary", "Online | " + event.title)
         cal_event.add("description", event.ingress_short)
         cal_event.add("uid", "event-" + str(event.id) + "@online.ntnu.no")
 


henrikhorluck:
Thank you for the contribution!

I think this is a feature better served by the user making their own iCal proxy that performs that modification?

I think users should be able to color-code their calendar in a way that it is obvious which events are Online-related, and this is a quite noisy change limiting how much information can be shown in notifications and on small screens

junlarsen:
> I think this is a feature better served by the user making their own iCal proxy that performs that modification?

I would argue this change is totally fine and I doubt a lot of people would be interested in running a proxy.

henrikhorluck:
> I would argue this change is totally fine and I doubt a lot of people would be interested in running a proxy.

My second point still stands, that this does take up a lot of valuable space with IMO more likely than not not-valuable information, that the user should have other tools available to fix.

I would on the other hand be open to us using e.g. an official Online-color as the default color on the calendar, which I think should be supported? (Though it only appears to maybe only support CSS3 color names) https://datatracker.ietf.org/doc/html/rfc7986#section-5.9

junlarsen:
A fixed color would be fine too imo.

henrikhorluck:
I'm closing this as I don't think we want _this_ exact change, but please propose a different option e.g. the online-color!

PR #3348, opened 2025-01-21
Add optimistic updates to checking in

jotjern:
Makes the client update instantly when checking in people to remove perceived slowness.


ErlendSae (approved):
Looks very good!

